PR #71, opened 2016-08-17
Remove deb/rpm profiles to prevent conflicts with CDAP

wolf31o2:
These profiles conflict with the `deb` and `rpm` profiles from `caskdata/cdap` when built using `caskdata/cdap-build` as a wrapper repository. The artifacts JARs from this repository are included in the `cdap-master` packages using `-Dsecurity.extensions.dir=...` and are available under `${CDAP_HOME}/master/ext/security` in the final packages.


bdmogal:
@wolf31o2 so after this, the recommended way of installing cdap-security-extensions is to install cdap-master (RPM/Parcel/Ambari), and then these would show up? Or just copy jar artifacts from the build to your cluster?

I think this is fine. However
1. Should we update the README with these instructions?
2. We could just remove the ``cdap-security-extensions-dist`` sub-module too now, right?

wolf31o2:
Correct. You would install `cdap-master` package on a package-based cluster. In a Parcel-based cluster, all files from a parcel are available on all cluster nodes just by adding the parcel to the cluster. Ambari re-uses the `cdap-master` package (and has no Sentry integration, so pretty much a moot point).

We should update the README, but let's get a solid build with this repository included to test before we update the docs.

As for removing the module, I tried that and didn't get any JARs built. I suspect we just need to move the pkg-prepare profile to the top-level POM, but this works as expected, when I tested it.

PR #72, opened 2016-08-17
Feature/remove action all

bdmogal:
Jira: [CDAP-6933](https://issues.cask.co/browse/CDAP-6933)
Build: http://builds.cask.co/browse/CDAP-CSI31, but for the build to pass, I'll need https://github.com/caskdata/cdap/pull/6473 to be merged, and a BUT to deploy the changes from it.

### DIFF
--- a/cdap-sentry/cdap-sentry-extension/cdap-sentry-binding/src/test/resources/test-authz-provider.ini
+++ b/cdap-sentry/cdap-sentry-extension/cdap-sentry-binding/src/test/resources/test-authz-provider.ini
@@ -25,28 +25,28 @@ readers2 = readers_ns2stream1
 ; namespace
 readers_ns1 = instance=cdap->namespace=ns1->action=read
 writers_ns1 = instance=cdap->namespace=ns1->action=write
-all_ns1 = instance=cdap->namespace=ns1->action=*
+all_ns1 = instance=cdap->namespace=ns1->action=all_hidden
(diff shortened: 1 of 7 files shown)


cdap-sentry/cdap-sentry-extension/cdap-sentry-binding/src/test/resources/test-authz-provider.ini
```diff
@@ -25,28 +25,28 @@ readers2 = readers_ns2stream1
 ; namespace
 readers_ns1 = instance=cdap->namespace=ns1->action=read
 writers_ns1 = instance=cdap->namespace=ns1->action=write
-all_ns1 = instance=cdap->namespace=ns1->action=*
+all_ns1 = instance=cdap->namespace=ns1->action=all_hidden
```
anwar6953: what is all_hidden

bdmogal: So we have this whole functionality in the sentry extension that makes "all" and "*" as special actions. When these are used, you're granted read,write,admin,execute. Although because of a recent CDAP PR you wouldn't be able to use ``ALL`` via CDAP, it will still work if you used Hue for example for managing privileges in Sentry. 

However, removing all that logic did not seem like a risk free change right now, so I put in this hack to not expose ALL as an action in the CDAP Sentry Extension.

bdmogal:
Build green:  http://builds.cask.co/browse/CDAP-CSI31

rohitsinha54:
LGTM but lets make sure we have a issue for removing ALL logic.

PR #73, opened 2016-08-18
Disable dataset based authorization extension since it creates a circ…

bdmogal:
…ular dependency. An entity on which authorization should be enforced (CDAP Dataset) shouldn't be the backend for authorization.


wolf31o2:
👍 LGTM

bdmogal:
Build: http://builds.cask.co/browse/CDAP-CSI32

wolf31o2:
👍 LGTM

PR #74, opened 2016-08-18
Switch to Apache Tephra

gokulavasan:


bdmogal:
LGTM, but I wonder why we depend on transitive dependencies, should we add a test dependency on Apache Tephra in this repo? Can do later.

bdmogal:
Let's merge this. I don't think adding a dependency is a great idea.

PR #76, opened 2016-08-19
Sentry service *requires* support for * as an action, so bringing it …

bdmogal:
…back


Nishith:
👍 LGTM.

bdmogal:
http://builds.cask.co/browse/CDAP-CSI34-1. Merging.

PR #89, opened 2016-12-14
CDAP-7680 Fix revoke on entity also used to revoke while deleting ent…

rohitsinha54:
…ities

Revoke on entities broke while removing support for ALL action in here https://github.com/caskdata/cdap-security-extn/pull/72 which was because the authorizatable were being added to the tAuthorizables 4 times (the number of actions we have) and equals was failing. 

This will also fix the issue where delete of entities was not deleting the privileges since delete entities calls revoke on entity. 

Issue: https://issues.cask.co/browse/CDAP-7680
Build: https://builds.cask.co/browse/CDAP-CSI37-1


anwar6953:
@Nishith can you be sure to enable the appropriate integration tests once this issue is fixed

Nishith:
Yes, will do.

sreevatsanraman:
LGTM

PR #90, opened 2016-12-17
fix compilation error by adding messaging methods to no op authorizat…

bdmogal:
…ion context


chtyim:
Ideally there should some test base provided by the auth extension api.

chtyim:
LGTM

bdmogal:
@chtyim yes, created https://issues.cask.co/browse/CDAP-7926.

PR #92, opened 2016-12-28
Implementing the new methods added to the AuthorizationContext

Nishith:


cdap-sentry/cdap-sentry-extension/cdap-sentry-binding/src/test/java/co/cask/cdap/security/authorization/sentry/binding/SentryAuthorizerTest.java
```diff
@@ -100,6 +102,32 @@ public Properties getExtensionProperties() {
       }
 
       @Override
+      public void createTopic(String s) throws TopicAlreadyExistsException, IOException {
```
anwar6953, line 105: Just a thought... isn't it weird to have such methods in a `AuthorizationContext` class?

anwar6953:
LGTM to unblock compilation.

cdap-sentry/cdap-sentry-extension/cdap-sentry-binding/src/test/java/co/cask/cdap/security/authorization/sentry/binding/SentryAuthorizerTest.java
Nishith: It is, but that is a TMS question.

anwar6953: For now, filed: https://issues.cask.co/browse/CDAP-7994

chtyim: It is not a TMS question. It is the authorization API issue. See my comment in CDAP-7994.

chtyim:
BTW, adding new method to api that requires changing code in this repo also reflect the issue in the auth API design. At the very least, it should provide test/mock classes, instead of requiring the extension implementation comes up with its own one.

PR #99, opened 2017-02-04
Update readme

rohitsinha54:
Specifically tell user to set authorization to enable as its needed for any other property to work
also change security.authorization.extension.config.instance.name to instance.name since in cdap we now have instance.name

### DIFF
--- a/cdap-sentry/cdap-sentry-extension/README.rst
+++ b/cdap-sentry/cdap-sentry-extension/README.rst
@@ -105,6 +105,9 @@ extension:
      * - Parameter
        - Description
        - Default Value
+     * - ``security.authorization.enabled``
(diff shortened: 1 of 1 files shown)


bdmogal:
0.4 is already released right? Do we plan on a 0.4.1? Otherwise merge to develop?

cdap-sentry/cdap-sentry-extension/README.rst
```diff
@@ -105,6 +105,9 @@ extension:
      * - Parameter
        - Description
        - Default Value
+     * - ``security.authorization.enabled``
```
bdmogal, line 108: can we describe this in a separate section outside this table? This table was supposed to contain only the properties that the security extension uses.

PR #105, opened 2017-03-18
Set grant option to true for all privileges in sentry and support getting roles for users

rohitsinha54:
Issue: https://issues.cask.co/browse/CDAP-9029 and https://issues.cask.co/browse/CDAP-9035

See issue for details

### DIFF
--- a/cdap-sentry/cdap-sentry-extension/cdap-sentry-binding/src/main/java/co/cask/cdap/security/authorization/sentry/binding/AuthBinding.java
+++ b/cdap-sentry/cdap-sentry-extension/cdap-sentry-binding/src/main/java/co/cask/cdap/security/authorization/sentry/binding/AuthBinding.java
@@ -69,6 +69,7 @@ import org.apache.sentry.provider.db.generic.SentryGenericProviderBackend;
 import org.apache.sentry.provider.db.generic.service.thrift.SentryGenericServiceClient;
 import org.apache.sentry.provider.db.generic.service.thrift.SentryGenericServiceClientFactory;
 import org.apache.sentry.provider.db.generic.service.thrift.TAuthorizable;
+import org.apache.sentry.provider.db.generic.service.thrift.TSentryGrantOption;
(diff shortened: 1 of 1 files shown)


cdap-sentry/cdap-sentry-extension/cdap-sentry-binding/src/main/java/co/cask/cdap/security/authorization/sentry/binding/AuthBinding.java
```diff
@@ -429,8 +430,24 @@ public Void run(SentryGenericServiceClient client) throws Exception {
...
+          return client.listAllRoles(requestingUser, COMPONENT_NAME);
+        }
+        if (principal.getType().equals(Principal.PrincipalType.USER)) {
+          // for a user get all the groups and their roles
+          Set<String> groups = authProvider.getGroupMapping().getGroups(principal.getName());
+          Set<TSentryRole> roles = new HashSet<>();
+          for (String group : groups) {
+            roles.addAll(client.listRolesByGroupName(requestingUser, group, COMPONENT_NAME));
+          }
+          return roles;
```
bdmogal: better to make this immutable?

bdmogal (approved):
1 comment about making the list of roles immutable. Anyway to add a test for these situations?

Other than that, LGTM

bdmogal:
anyway we can add a test for this?

rohitsinha54:
There is a unit test which tests the model policy part from  static file. I think we investigating it before but you cannot start a db based Sentry which allows revoking/granting operation. The CDAP Authorization integration test is supposed to test all this not sure what it does. 

Although its very clear now that we need all these tests. We can work on writing an authorization test which does this.

rohitsinha54:
@bdmogal : Merged the PR and have opened a jira to add tests for RBAC scenario as a blocker for 4.2 https://issues.cask.co/browse/CDAP-9047

PR #108, opened 2017-03-23
CDAP-9027 support grant and revoke on user and group

rohitsinha54:
- Added support for grant to groups
- Added support for revoke to groups
- Added support for revoke to users

Support for grant to user was already existing 

Issue: https://issues.cask.co/browse/CDAP-9027

### DIFF
--- a/cdap-sentry/cdap-sentry-extension/cdap-sentry-binding/src/main/java/co/cask/cdap/security/authorization/sentry/binding/SentryAuthorizer.java
+++ b/cdap-sentry/cdap-sentry-extension/cdap-sentry-binding/src/main/java/co/cask/cdap/security/authorization/sentry/binding/SentryAuthorizer.java
@@ -78,22 +78,43 @@ public class SentryAuthorizer extends AbstractAuthorizer {
         binding.grant(entityId, new Role(principal.getName()), actions, getRequestingUser());
         break;
       case USER:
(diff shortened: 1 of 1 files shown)


cdap-sentry/cdap-sentry-extension/cdap-sentry-binding/src/main/java/co/cask/cdap/security/authorization/sentry/binding/SentryAuthorizer.java
```diff
@@ -78,22 +78,41 @@ public void grant(EntityId entityId, Principal principal, Set<Action> actions) t
...
+        break;
+      case USER:
+        // get the group for the user and the role associated with entity and perform revoke
+        Role entityRole = getEntityUserRole(entityId, getGroupPrincipal(principal));
+        binding.revoke(entityId, entityRole, actions, getRequestingUser());
+        cleanUpRole(entityRole);
+        break;
+      case GROUP:
+        // get the role associated with the entity for the group and perform revoke
+        binding.revoke(entityId, getEntityUserRole(entityId, principal), actions, getRequestingUser());
```
Nishith: getEntityUserRole is called twice, better to create a variable.

Nishith (approved):
One minor comment, LGTM

PR #113, opened 2017-04-14
Bumping up version

Nishith:


cdap-authorization-dataset-extn/pom.xml
```diff
@@ -22,7 +22,7 @@
   <parent>
     <groupId>co.cask.cdap</groupId>
     <artifactId>cdap-security-extensions</artifactId>
-    <version>0.4.1</version>
+    <version>0.4.2</version>
```
sreevatsanraman: All the versions should be snapshot

Nishith: I thought we were going to release this with cdap. Will change to -SNAPSHOT.

Nishith:
Updated the version. Please take another look.

PR #118, opened 2017-05-19
[CDAP-11366] Adding a cache to the Sentry authorizer implementation.

Nishith:
The extension now caches enforcement calls. The cache properties can be set by using the "security.authorization.extension.config." prefix in cdap-site. The complimentary PR in CDAP adding those configuration options will be opened after https://github.com/caskdata/cdap/pull/8750 is merged.

### DIFF
--- a/cdap-sentry/cdap-sentry-extension/cdap-sentry-binding/src/main/java/co/cask/cdap/security/authorization/sentry/binding/SentryAuthorizer.java
+++ b/cdap-sentry/cdap-sentry-extension/cdap-sentry-binding/src/main/java/co/cask/cdap/security/authorization/sentry/binding/SentryAuthorizer.java
@@ -132,7 +132,6 @@ public class SentryAuthorizer extends AbstractAuthorizer {
           String.format("The given principal '%s' is of unsupported type '%s'.", principal.getName(),
                         principal.getType()));
     }
-    loadCache(entityId, principal, actions, true);
     LOG.trace("Granted {} on {} to {}", actions, entityId, principal);
   }
 
@@ -165,7 +164,6 @@ public class SentryAuthorizer extends AbstractAuthorizer {
           String.format("The given principal '%s' is of unsupported type '%s'.", principal.getName(),
                         principal.getType()));
     }
-    loadCache(entityId, principal, actions, false);
     LOG.trace("Revoked {} on {} to {}", actions, entityId, principal);
   }
 
@@ -258,13 +256,6 @@ public class SentryAuthorizer extends AbstractAuthorizer {
     for (Action action : actions) {
       AuthorizationPrivilege authorizationPrivilege = new AuthorizationPrivilege(principal, entityId, action);
       authPolicyCache.invalidate(authorizationPrivilege);
-    }
-  }
-
-  private void loadCache(EntityId entityId, Principal principal, Set<Action> actions, Boolean allowed) {
-    for (Action action : actions) {
-      AuthorizationPrivilege authorizationPrivilege = new AuthorizationPrivilege(principal, entityId, action);
-      authPolicyCache.put(authorizationPrivilege, allowed);
     }
   }
 
--- a/cdap-sentry/cdap-sentry-extension/cdap-sentry-binding/src/main/java/co/cask/cdap/security/authorization/sentry/binding/SentryAuthorizer.java
+++ b/cdap-sentry/cdap-sentry-extension/cdap-sentry-binding/src/main/java/co/cask/cdap/security/authorization/sentry/binding/SentryAuthorizer.java
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2016 Cask Data, Inc.
+ * Copyright © 2016-2017 Cask Data, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not
  * use this file except in compliance with the License. You may obtain a copy of
@@ -28,18 +28,26 @@ import co.cask.cdap.security.spi.authorization.Authorizer;
 import co.cask.cdap.security.spi.authorization.RoleAlreadyExistsException;
 import co.cask.cdap.security.spi.authorization.RoleNotFoundException;
 import co.cask.cdap.security.spi.authorization.UnauthorizedException;
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Joiner;
 import com.google.common.base.Preconditions;
 import com.google.common.base.Predicate;
 import com.google.common.base.Strings;
+import com.google.common.cache.CacheBuilder;
+import com.google.common.cache.CacheLoader;
+import com.google.common.cache.LoadingCache;
 import com.google.common.collect.Collections2;
 import com.google.common.collect.Sets;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import java.util.HashSet;
+import java.util.Collections;
+import java.util.EnumSet;
+import java.util.Map;
+import java.util.Objects;
 import java.util.Properties;
 import java.util.Set;
+import java.util.concurrent.TimeUnit;
 
 /**
  * This class implements {@link Authorizer} from CDAP and is responsible for interacting with Sentry to manage
@@ -49,6 +57,14 @@ public class SentryAuthorizer extends AbstractAuthorizer {
 
   private static final Logger LOG = LoggerFactory.getLogger(SentryAuthorizer.class);
   private static final String ENTITY_ROLE_PREFIX = ".";
+
+  // Cache for privileges. It stores the result of an enforce call.
+  // [Principal, Entity, Action] -> Boolean
+  // Both positive and negative results are stored. The cache entry is added/invalidated
+  // in case of a grant/revoke call for a principal.
+  private static LoadingCache<AuthorizationPrivilege, Boolean> authPolicyCache;
+
+  private boolean cacheEnabled;
   private AuthBinding binding;
   private AuthorizationContext context;
 
@@ -69,6 +85,28 @@ public class SentryAuthorizer extends AbstractAuthorizer {
       properties.getProperty(AuthConf.INSTANCE_NAME) :
       AuthConf.AuthzConfVars.getDefault(AuthConf.INSTANCE_NAME);
 
+    int cacheTtlSecs = Integer.parseInt(properties.getProperty(AuthConf.CACHE_TTL_SECS,
+                                                               AuthConf.CACHE_TTL_SECS_DEFAULT));
+    int cacheMaxEntries = Integer.parseInt(properties.getProperty(AuthConf.CACHE_MAX_ENTRIES,
+                                                                  AuthConf.CACHE_MAX_ENTRIES_DEFAULT));
+    this.cacheEnabled = cacheMaxEntries > 0;
+
+    if (cacheEnabled) {
+      authPolicyCache = CacheBuilder.newBuilder()
+        .expireAfterWrite(cacheTtlSecs, TimeUnit.SECONDS)
+        .maximumSize(cacheMaxEntries)
+        .build(new CacheLoader<AuthorizationPrivilege, Boolean>() {
+          @SuppressWarnings("NullableProblems")
+          @Override
+          public Boolean load(AuthorizationPrivilege authorizationPrivilege) throws Exception {
+            LOG.trace("Cache miss for {}", authorizationPrivilege);
+            return doEnforce(authorizationPrivilege);
+          }
+        });
+    } else {
+      authPolicyCache = null;
+    }
+
     LOG.info("Configuring SentryAuthorizer with sentry-site.xml at {}, CDAP instance {} and Sentry Admin Group: {}",
                sentrySiteUrl, instanceName, sentryAdminGroup);
     this.binding = new AuthBinding(sentrySiteUrl, instanceName, sentryAdminGroup);
@@ -77,6 +115,7 @@ public class SentryAuthorizer extends AbstractAuthorizer {
 
   @Override
   public void grant(EntityId entityId, Principal principal, Set<Action> actions) throws RoleNotFoundException {
+    invalidateCache(entityId, principal, actions);
     switch (principal.getType()) {
       case ROLE:
         binding.grant(entityId, new Role(principal.getName()), actions, getRequestingUser());
@@ -93,10 +132,13 @@ public class SentryAuthorizer extends AbstractAuthorizer {
           String.format("The given principal '%s' is of unsupported type '%s'.", principal.getName(),
                         principal.getType()));
     }
+    loadCache(entityId, principal, actions, true);
+    LOG.trace("Granted {} on {} to {}", actions, entityId, principal);
   }
 
   @Override
   public void revoke(EntityId entityId, Principal principal, Set<Action> actions) throws RoleNotFoundException {
+    invalidateCache(entityId, principal, actions);
     Role entityRole;
     switch (principal.getType()) {
       case ROLE:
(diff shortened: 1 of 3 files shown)


Nishith:
Adding unit tests will update PR with that.

rohitsinha54:
Please add unit tests in a separate commit so it's easier to review later.

Nishith:
Unit tests are already in a separate commit. Do you mean in a separate PR?

cdap-sentry/cdap-sentry-extension/cdap-sentry-binding/src/main/java/co/cask/cdap/security/authorization/sentry/binding/SentryAuthorizer.java
```diff
@@ -49,6 +57,10 @@
 
   private static final Logger LOG = LoggerFactory.getLogger(SentryAuthorizer.class);
   private static final String ENTITY_ROLE_PREFIX = ".";
+
+  private static LoadingCache<AuthorizationPrivilege, Boolean> authPolicyCache;
```
rohitsinha54: a comment on what is this cache and what it does with other relevant details

cdap-sentry/cdap-sentry-extension/cdap-sentry-binding/src/main/java/co/cask/cdap/security/authorization/sentry/binding/SentryAuthorizer.java
```diff
@@ -69,6 +81,26 @@ public void initialize(AuthorizationContext context) throws Exception {
       properties.getProperty(AuthConf.INSTANCE_NAME) :
       AuthConf.AuthzConfVars.getDefault(AuthConf.INSTANCE_NAME);
 
+    int cacheTtlSecs = Integer.parseInt(properties.getProperty(AuthConf.CACHE_TTL_SECS));
```
rohitsinha54: it looks like properties.getProperty(AuthConf.CACHE_TTL_SECS) and the below can be null if not specified the parseInt will throw NumberFormatException  correct ?

cdap-sentry/cdap-sentry-extension/cdap-sentry-binding/src/main/java/co/cask/cdap/security/authorization/sentry/binding/SentryAuthorizer.java
```diff
@@ -77,6 +109,7 @@ public void initialize(AuthorizationContext context) throws Exception {
 
   @Override
   public void grant(EntityId entityId, Principal principal, Set<Action> actions) throws RoleNotFoundException {
+    LOG.debug("Granting {} on {} to {}", actions, entityId, principal);
```
rohitsinha54: if i remember correctly you added log for grant/revoke etc with info level in AuthorizationHandler or some class which is called before this. SentryAuthorizer runs inside master too so it will lead to multiple logging. 

Also the binding.grant (which is in master too) call logs too so might not be needed here

cdap-sentry/cdap-sentry-extension/cdap-sentry-binding/src/main/java/co/cask/cdap/security/authorization/sentry/binding/SentryAuthorizer.java
```diff
@@ -278,4 +331,66 @@ private String getRequestingUser() throws IllegalArgumentException {
     LOG.trace("Got requesting principal {}", principal);
     return principal.getName();
   }
+
+  @VisibleForTesting
+  public Map<AuthorizationPrivilege, Boolean> cacheAsMap() {
```
rohitsinha54, line 357: so this needs to be done everywhere we will create the cache ?  i saw this class in cdap code base too. I dont think extensions should be defining this class. can we keep it in one place in cdap

cdap-sentry/cdap-sentry-extension/cdap-sentry-binding/src/main/java/co/cask/cdap/security/authorization/sentry/binding/conf/AuthConf.java
```diff
@@ -37,6 +37,11 @@
   public static final String INSTANCE_NAME = "instance.name";
   // a comma separated list of users who will be superusers
   public static final String SUPERUSERS = "superusers";
+  // TTL for authorization cache entries
+  public static final String CACHE_TTL_SECS = "cache.ttl.secs";
+  // Maximum number of entries to keep in the cache. An entry of <= 0 would disable caching.
```
rohitsinha54: is this a good idea ? I am not sure. Probably ask Andreas or Terence. Should we explicitly have a  flag to enable/disable ?

cdap-sentry/cdap-sentry-extension/cdap-sentry-binding/src/test/java/co/cask/cdap/security/authorization/sentry/binding/SentryAuthorizerTest.java
```diff
@@ -272,6 +278,46 @@ public void testHierarchy() throws Exception {
...
+  @Test
+  public void testCache() throws Exception {
+    final NamespaceId ns1 = new NamespaceId("ns1");
+    final Principal alice = getUser("alice");
+    final Principal admin1 = getUser("admin1");
+    try {
+      assertUnauthorized(ns1, alice, Action.READ);
+      assertAuthorized(ns1, admin1, Action.ADMIN);
+    } catch (UnauthorizedException ignored) {
+      //
```
rohitsinha54: nit: // expected

cdap-sentry/cdap-sentry-extension/cdap-sentry-binding/src/test/java/co/cask/cdap/security/authorization/sentry/binding/SentryAuthorizerTest.java
```diff
@@ -272,6 +278,46 @@ public void testHierarchy() throws Exception {
...
+    }
+    Assert.assertTrue(2 == authorizer.cacheAsMap().size());
```
rohitsinha54: how is this 2 ? because of the unauthorized which makes the entry with false value ?

cdap-sentry/cdap-sentry-extension/cdap-sentry-binding/src/main/java/co/cask/cdap/security/authorization/sentry/binding/SentryAuthorizer.java
```diff
@@ -180,9 +218,24 @@ public void enforce(EntityId entityId, Principal principal, Set<Action> actions)
     Preconditions.checkArgument(Principal.PrincipalType.USER == principal.getType(), "The given principal '%s' is of " +
                                 "type '%s'. In Sentry authorization checks can only be performed on principal type " +
                                 "'%s'.", principal.getName(), principal.getType(), Principal.PrincipalType.USER);
-    if (!binding.authorize(entityId, principal, actions)) {
-      throw new UnauthorizedException(principal, actions, entityId);
+    Set<Action> disallowed = EnumSet.noneOf(Action.class);
```
rohitsinha54, line 235: How does this cache works ? On enforce its loaded with the AuthorizationPrivilege and if the authorization is not  there the boolean says false. Now when you do grant don't you need to refresh/invalidate the cache else the cache will stay say unauthorized ?  same for revoke.

cdap-sentry/cdap-sentry-extension/cdap-sentry-binding/src/test/java/co/cask/cdap/security/authorization/sentry/binding/SentryAuthorizerTest.java
```diff
@@ -272,6 +278,46 @@ public void testHierarchy() throws Exception {
...
+      assertUnauthorized(ns1, admin1, Action.READ);
+    } catch (UnauthorizedException ignored) {
+      //
+    }
+    Assert.assertTrue(authorizer.cacheAsMap().containsKey(new SentryAuthorizer.AuthorizationPrivilege(admin1, ns1,
+                                                                                                       Action.READ)));
+    Assert.assertFalse(authorizer.cacheAsMap().containsKey(new SentryAuthorizer.AuthorizationPrivilege(admin1, ns1,
+                                                                                                      Action.ADMIN)));
+    Assert.assertFalse(authorizer.cacheAsMap().containsKey(new SentryAuthorizer.AuthorizationPrivilege(admin1, ns1,
+                                                                                                       Action.WRITE)));
```
rohitsinha54, line 318: should test the cache update on grant too ? Although might not be possible since this follows the file based storage and grant can't be done correct ?

rohitsinha54 (changes requested):

rohitsinha54:
@Nishith: Yes, I think when I reviewed the unit tests were already in. 👍  I have added some comments on the changes. Please  have a look.

cdap-sentry/cdap-sentry-extension/cdap-sentry-binding/src/main/java/co/cask/cdap/security/authorization/sentry/binding/SentryAuthorizer.java
Nishith: This needs to be where ever the cache is because different extensions can have their own caching scheme. In this case, they happen to be the same but in general, cdap should not be dictating how the extensions should cache.

cdap-sentry/cdap-sentry-extension/cdap-sentry-binding/src/main/java/co/cask/cdap/security/authorization/sentry/binding/conf/AuthConf.java
Nishith: This is according to Terence's suggestion here https://github.com/caskdata/cdap/pull/8750#discussion-diff-117072821R2719

cdap-sentry/cdap-sentry-extension/cdap-sentry-binding/src/test/java/co/cask/cdap/security/authorization/sentry/binding/SentryAuthorizerTest.java
Nishith: Yes, we cache the false value as well.

cdap-sentry/cdap-sentry-extension/cdap-sentry-binding/src/test/java/co/cask/cdap/security/authorization/sentry/binding/SentryAuthorizerTest.java
Nishith: This case should be covered by an existing integration test. **https://github.com/caskdata/cdap-integration-tests/blob/develop/integration-test-remote/src/test/java/co/cask/cdap/security/BasicSecurityTest.java#L105**

cdap-sentry/cdap-sentry-extension/cdap-sentry-binding/src/main/java/co/cask/cdap/security/authorization/sentry/binding/SentryAuthorizer.java
Nishith: It is generally helpful to find out the place where the call flow broke in case of an error so I will change it to trace but leave it here.

cdap-sentry/cdap-sentry-extension/cdap-sentry-binding/src/main/java/co/cask/cdap/security/authorization/sentry/binding/SentryAuthorizer.java
```diff
@@ -238,6 +245,20 @@ private Boolean doEnforce(AuthorizationPrivilege authorizationPrivilege) {
                       Collections.singleton(authorizationPrivilege.getAction()));
   }
 
+  private void invalidateCache(EntityId entityId, Principal principal, Set<Action> actions) {
```
rohitsinha54: The cache needs to be invalidated when an entity is deleted too else in scenario where entity is deleted and created again with same name it will have the privilege for users in cache.

cdap-sentry/cdap-sentry-extension/cdap-sentry-binding/src/test/java/co/cask/cdap/security/authorization/sentry/binding/SentryAuthorizerTest.java
```diff
@@ -72,7 +72,7 @@ public SentryAuthorizerTest() throws Exception {
     properties.put(AuthConf.INSTANCE_NAME, "cdap");
     properties.put(AuthConf.SUPERUSERS, Joiner.on(",").join(SUPERUSER_HULK, SUPERUSER_SPIDERMAN));
     properties.put(AuthConf.SENTRY_ADMIN_GROUP, "cdap");
-    properties.put(AuthConf.CACHE_MAX_ENTRIES, "100");
+//    properties.put(AuthConf.CACHE_MAX_ENTRIES, "100");
```
rohitsinha54: commented line of code ?

cdap-sentry/cdap-sentry-extension/cdap-sentry-binding/src/main/java/co/cask/cdap/security/authorization/sentry/binding/SentryAuthorizer.java
rohitsinha54: But can we keep this AuthorizationPrivilege class in cdap.security spi or whatever we expose for commons/util in this way we will not have code duplication. extensions which want to use can use it doesn't seem like an internal thing to not expose it

cdap-sentry/cdap-sentry-extension/cdap-sentry-binding/src/test/java/co/cask/cdap/security/authorization/sentry/binding/SentryAuthorizerTest.java
rohitsinha54: that test  does not test the cache invalidation on grant/revoke which should reside here but i guess its not achievable as of now in unit test  because of file based policy management here. so its fine. ignore the comment.

rohitsinha54:
@Nishith : Few comments mostly are minor but one piece which is missing is cache invalidation during entity deletion.

cdap-sentry/cdap-sentry-extension/cdap-sentry-binding/src/main/java/co/cask/cdap/security/authorization/sentry/binding/SentryAuthorizer.java
Nishith: The cache scheme for the container _is_ internal. If we push this class to the SPI or any place external then either it would be really difficult to change that or we will have to create a new class for that key. We are already talking about changing the cache scheme based on the type of container. So, this class will very likely change, that will become very difficult to do if it was part of the SPI. Sentry also has client-side caching on their roadmap, if and when that becomes available we might not even need the cache here and then we will end up with a class polluting the SPI that no one is using.

cdap-sentry/cdap-sentry-extension/cdap-sentry-binding/src/main/java/co/cask/cdap/security/authorization/sentry/binding/SentryAuthorizer.java
```diff
@@ -123,17 +165,29 @@ public void revoke(EntityId entityId, Principal principal, Set<Action> actions)
           String.format("The given principal '%s' is of unsupported type '%s'.", principal.getName(),
                         principal.getType()));
     }
+    loadCache(entityId, principal, actions, false);
```
rohitsinha54: Why load the cache on grant/revoke ? Do we expect enforce call to come right after in every scenario ? I think the cache should only be loaded at one place i.e. enforce. These guys should only invalidate.

rohitsinha54 (commented):
One comment other than that LGTM

rohitsinha54:
Please test on cluster before merging.

cdap-sentry/cdap-sentry-extension/cdap-sentry-binding/src/main/java/co/cask/cdap/security/authorization/sentry/binding/SentryAuthorizer.java
anwar6953: done.

anwar6953:
We'll test this change after merging (so as to test this change and all the other authorization changes in one pass).

rohitsinha54:
LGTM. Please ensure that the changes work.

PR #123, opened 2017-06-06
CDAP-11829 Change cache scheme. Load privileges in batch upon cache miss.

anwar6953:
CDAP-11829 Change cache scheme. Load privileges in batch upon cache miss.

https://issues.cask.co/browse/CDAP-11829

PR #124, opened 2017-06-06
Revert "CDAP-11792 Do not propagate instance privileges"

anwar6953:
Reverts caskdata/cdap-security-extn#120

This change is no longer needed as the implementation of enforcing privileges has changed in https://github.com/caskdata/cdap-security-extn/pull/123, such that we no longer rely on Sentry's `hasAccess`, which was allowing access due to hierarchical privileges.

Reverting this change also removes the need for an upgrade step.

### DIFF
--- a/cdap-sentry/cdap-sentry-extension/cdap-sentry-binding/src/main/java/co/cask/cdap/security/authorization/sentry/binding/AuthBinding.java
+++ b/cdap-sentry/cdap-sentry-extension/cdap-sentry-binding/src/main/java/co/cask/cdap/security/authorization/sentry/binding/AuthBinding.java
@@ -701,7 +701,7 @@ class AuthBinding {
         authorizables.add(new Instance(((InstanceId) entityId).getInstance()));
         break;
       case NAMESPACE:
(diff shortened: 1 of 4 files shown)


rohitsinha54:
I think we should still keep this change to keep the behavior consistent with the CDAP. 
@poornachandra : Thoughts ?

anwar6953:
What do you mean by "keep the behavior consistent with the CDAP"?
How will the behavior be different if we revert this change?

rohitsinha54:
CDAP enforcement / privilege propagation behaves in a way that the privileges on instance is not propagated to namespace.  Since InstanceId is not parent of ParentId. Behavior will be different if some other client (say like hue or something else) instantiates SentryAuthorizer and uses it I think. Even we do this in some of our external tools.

anwar6953:
The same will be true for this extension. See the original comment for why pull #123 brings back that behavior without the need for the change in pull #120.

poornachandra (approved):

rohitsinha54:
👍

PR #131, opened 2017-07-17
CDAP-12135 Change caching model to reduce memory usage by avoiding denormalized caching

poornachandra:
The idea is to replace one big cache of `Principal, Entity, Action] -> Boolean` with three smaller caches - group cache, role cache and entity-actions cache. This is so that we avoid caching duplicate entries (entity/aciton) for every principal belonging to the same role.

Note: the entity-actions cache will be replaced by a wildcard policy cache in the up-coming PR.

Build - https://builds.cask.co/browse/CDAP-CSI58-5

### DIFF
--- a/cdap-sentry/cdap-sentry-extension/cdap-sentry-binding/src/main/java/co/cask/cdap/security/authorization/sentry/binding/AuthBinding.java
+++ b/cdap-sentry/cdap-sentry-extension/cdap-sentry-binding/src/main/java/co/cask/cdap/security/authorization/sentry/binding/AuthBinding.java
@@ -50,14 +50,14 @@ import co.cask.cdap.security.authorization.sentry.policy.PrivilegeValidator;
 import co.cask.cdap.security.spi.authorization.AlreadyExistsException;
 import co.cask.cdap.security.spi.authorization.BadRequestException;
 import co.cask.cdap.security.spi.authorization.NotFoundException;
-import co.cask.cdap.security.spi.authorization.RoleAlreadyExistsException;
-import co.cask.cdap.security.spi.authorization.RoleNotFoundException;
 import co.cask.cdap.security.spi.authorization.UnauthorizedException;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
 import com.google.common.base.Strings;
 import com.google.common.base.Throwables;
-import com.google.common.collect.ImmutableList;
+import com.google.common.cache.CacheBuilder;
+import com.google.common.cache.CacheLoader;
+import com.google.common.cache.LoadingCache;
 import com.google.common.collect.ImmutableSet;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.sentry.policy.common.PolicyEngine;
@@ -65,7 +65,6 @@ import org.apache.sentry.provider.common.AuthorizationProvider;
 import org.apache.sentry.provider.common.ProviderBackend;
 import org.apache.sentry.provider.db.SentryAccessDeniedException;
 import org.apache.sentry.provider.db.SentryAlreadyExistsException;
-import org.apache.sentry.provider.db.SentryGrantDeniedException;
 import org.apache.sentry.provider.db.SentryInvalidInputException;
 import org.apache.sentry.provider.db.SentryNoSuchObjectException;
 import org.apache.sentry.provider.db.SentryThriftAPIMismatchException;
@@ -83,11 +82,16 @@ import java.lang.reflect.Constructor;
 import java.net.MalformedURLException;
 import java.net.URL;
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.Collections;
+import java.util.EnumSet;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.LinkedList;
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.TimeUnit;
 import javax.annotation.Nullable;
 
 /**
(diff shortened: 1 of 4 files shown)


cdap-sentry/cdap-sentry-extension/cdap-sentry-binding/src/main/java/co/cask/cdap/security/authorization/sentry/binding/AuthBinding.java
```diff
@@ -103,11 +107,111 @@
   private final String instanceName;
   private final String sentryAdminGroup;
 
-  AuthBinding(String sentrySite, String instanceName, String sentryAdminGroup) {
+  private final LoadingCache<Principal, Set<String>> groupCache;
+  private final LoadingCache<String, Set<Role>> roleCache;
```
yaojiefeng, line 111: It would be good to add some comments about what is the cache key and cache value. It is a bit unclear to me when I first see the string type cache key.

Also this cache needs to invalidated for a user if addRole, dropRole etc happens

cdap-sentry/cdap-sentry-extension/cdap-sentry-binding/src/main/java/co/cask/cdap/security/authorization/sentry/binding/AuthBinding.java
```diff
@@ -103,11 +107,111 @@
...
+          return fetchEntityActions(role);
+        }
+      });
+  }
+
+  private Set<String> getGroups(Principal principal) {
+    return authProvider.getGroupMapping().getGroups(principal.getName());
+  }
+
+  private Set<Role> getRoles(final String group) throws Exception {
```
yaojiefeng, line 162: There is another `getRoles` method at Line 571, which used to get the roles for any kind of principal. Is it better to call this method `getRolesForGroup` to avoid confusion?

cdap-sentry/cdap-sentry-extension/cdap-sentry-binding/src/main/java/co/cask/cdap/security/authorization/sentry/binding/AuthBinding.java
```diff
@@ -638,6 +731,7 @@ protected void revoke(final EntityId entityId, final Role role, Set<Action> acti
   }
 
   private SentryGenericServiceClient getClient() throws Exception {
+    // TODO: Cache the client
```
yaojiefeng, line 734: I am not sure I understand this method. Can you just create the client in the constructor and have it as a global variable since I do not see it will change in the code.

yaojiefeng (commented):
Just some small comments. Also, do we need some unit tests for the new caches?

cdap-sentry/cdap-sentry-extension/cdap-sentry-binding/src/main/java/co/cask/cdap/security/authorization/sentry/binding/AuthBinding.java
poornachandra: Thrift client is not thread-safe, and hence we cannot create a single instance

poornachandra:
@yaojiefeng The existing unit tests cover the testing of the newly introduced caches.

cdap-sentry/cdap-sentry-extension/cdap-sentry-binding/src/main/java/co/cask/cdap/security/authorization/sentry/binding/AuthBinding.java
poornachandra: This method is renamed to `fetchRoles()` in the upcoming PR

cdap-sentry/cdap-sentry-extension/cdap-sentry-binding/src/main/java/co/cask/cdap/security/authorization/sentry/binding/AuthBinding.java
poornachandra: Will add comments

cdap-sentry/cdap-sentry-extension/cdap-sentry-binding/src/main/java/co/cask/cdap/security/authorization/sentry/binding/AuthBinding.java
```diff
@@ -103,11 +107,111 @@
   private final String instanceName;
   private final String sentryAdminGroup;
 
-  AuthBinding(String sentrySite, String instanceName, String sentryAdminGroup) {
+  private final LoadingCache<Principal, Set<String>> groupCache;
```
rohitsinha54, line 110: Do we really need a cache for user to group ? I don't think it's an expensive call. Furthermore looking into the getGroups implementation I see that it already has a cache internally which it uses first. Adding another cache here will just cause overhead and can lead to inconsistencies. Any specific reason to add it here ?

cdap-sentry/cdap-sentry-extension/cdap-sentry-binding/src/main/java/co/cask/cdap/security/authorization/sentry/binding/AuthBinding.java
```diff
@@ -103,11 +107,111 @@
...
+        @SuppressWarnings("NullableProblems")
+        @Override
+        public Map<EntityId, Set<Action>> load(final Role role) throws Exception {
+          LOG.trace("Privilege cache miss for role {}", role);
+          return fetchEntityActions(role);
+        }
+      });
+  }
+
+  private Set<String> getGroups(Principal principal) {
```
rohitsinha54, line 158: nit: These private method should go lower in the file according to the visibility.

cdap-sentry/cdap-sentry-extension/cdap-sentry-binding/src/main/java/co/cask/cdap/security/authorization/sentry/binding/AuthBinding.java
```diff
@@ -103,11 +107,111 @@
...
+  private Map<EntityId, Set<Action>> fetchEntityActions(final Role role) throws Exception {
+    Set<TSentryPrivilege> sentryPrivileges = execute(new Command<Set<TSentryPrivilege>>() {
+      @Override
+      public Set<TSentryPrivilege> run(SentryGenericServiceClient client) throws Exception {
+        return client.listPrivilegesByRoleName(sentryAdminGroup, role.getName(), COMPONENT_NAME, instanceName);
+      }
+    });
+    Set<Privilege> privileges = toPrivileges(sentryPrivileges);
+
+    if (privileges == null) {
```
rohitsinha54, line 185: This can never be bull since we do Set<Privilege> privileges = new HashSet<>(); in toPrivileges. I think your intention was to check for sentryPrivileges ? Although, I think that will also be empty list if no privilege exists rather than null.

cdap-sentry/cdap-sentry-extension/cdap-sentry-binding/src/test/java/co/cask/cdap/security/authorization/sentry/binding/SentryAuthorizerTest.java
```diff
@@ -275,30 +275,6 @@ public void testUnauthorized() throws Exception {
     assertUnauthorized(new StreamId("ns1", "stream1"), getUser("admin1"), Action.EXECUTE);
   }
 
-  @Test
-  public void testCache() throws Exception {
```
rohitsinha54, line 279: We should have some test for the caches we added too

cdap-sentry/cdap-sentry-extension/cdap-sentry-binding/src/main/java/co/cask/cdap/security/authorization/sentry/binding/AuthBinding.java
```diff
@@ -103,11 +107,111 @@
...
+        actions = EnumSet.noneOf(Action.class);
+        result.put(privilege.getEntity(), actions);
+      }
+      actions.add(privilege.getAction());
+    }
+    LOG.debug("Got entity-actions {} for role {}", result, role);
+    return result;
+  }
+
+  Set<Action> getActions(Principal principal, EntityId entityId) throws Exception {
```
rohitsinha54, line 203: should we call getAllowedActions or something just for clarity

rohitsinha54 (commented):
Took a pass. Few comments.

cdap-sentry/cdap-sentry-extension/cdap-sentry-binding/src/main/java/co/cask/cdap/security/authorization/sentry/binding/AuthBinding.java
poornachandra: The group lookup happens on every enforce call. The group mapper implementation can be replaced with a custom implementation which may not have a cache.

Since these new caches will be read-only and won't have any updates to them from the authorizer end (once we remove grants/revokes), I don't see any reason for inconsistency. Also not sure what overhead this cache will add other than a few extra KB of memory used. Let me know if have any other inconsistency or overhead in mind.

cdap-sentry/cdap-sentry-extension/cdap-sentry-binding/src/main/java/co/cask/cdap/security/authorization/sentry/binding/AuthBinding.java
poornachandra: This method will be removed in the next PR

cdap-sentry/cdap-sentry-extension/cdap-sentry-binding/src/main/java/co/cask/cdap/security/authorization/sentry/binding/AuthBinding.java
poornachandra: This method will also be removed in the next PR

cdap-sentry/cdap-sentry-extension/cdap-sentry-binding/src/main/java/co/cask/cdap/security/authorization/sentry/binding/AuthBinding.java
poornachandra: I'll move them out in the upcoming PR.

cdap-sentry/cdap-sentry-extension/cdap-sentry-binding/src/test/java/co/cask/cdap/security/authorization/sentry/binding/SentryAuthorizerTest.java
poornachandra: This test was just testing the eviction since we were using a custom weight function for eviction. The existing unit tests will test the caches. However, I plan to add performance tests to test the efficiency of the caches later.

cdap-sentry/cdap-sentry-extension/cdap-sentry-binding/src/main/java/co/cask/cdap/security/authorization/sentry/binding/AuthBinding.java
rohitsinha54: I don't think that the implementation will change to not a have a cache if it already has one I think we can rely on it. But if it really seems something which we want to be certain of then we can have the cache. 

With inconsistency and overhead I meant what happens with this cache when an user is added or removed from a group in the underlying system. Since the call does not pass through CDAP it has no way to invalidate/refresh the entry so we will have to wait for it to be refreshed on time duration.

cdap-sentry/cdap-sentry-extension/cdap-sentry-binding/src/main/java/co/cask/cdap/security/authorization/sentry/binding/AuthBinding.java
rohitsinha54: 👍

cdap-sentry/cdap-sentry-extension/cdap-sentry-binding/src/main/java/co/cask/cdap/security/authorization/sentry/binding/AuthBinding.java
poornachandra: What I meant is that the implementation can change to use LDAP lookup or something. 

Also since we are removing grants/revokes, we have to wait for the cache timeout for any change - group, role or privilege. We could reduce the timeout for cache on the master side if required, can you file a JIRA for this?

rohitsinha54: https://issues.cask.co/browse/CDAP-12144

rohitsinha54:
LGTM

poornachandra:
Build is green, merging...

PR #132, opened 2017-07-19
CDAP-12148 Initial Integration with Apache Ranger

rohitsinha54:
Integration with Apache Ranger

This PR includes the following
- A binding which is responsible for lookup
- A lookup which performs test connection
- Unit test will be added in another commit/PR

### DIFF
--- a/cdap-ranger/cdap-ranger-binding/src/main/java/co/cask/cdap/security/authorization/ranger/binding/RangerAuthorizer.java
+++ b/cdap-ranger/cdap-ranger-binding/src/main/java/co/cask/cdap/security/authorization/ranger/binding/RangerAuthorizer.java
@@ -1,0 +1,298 @@
+/*
+ * Copyright © 2017 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package co.cask.cdap.security.authorization.ranger.binding;
+
+import co.cask.cdap.common.conf.Constants;
+import co.cask.cdap.proto.element.EntityType;
+import co.cask.cdap.proto.id.ApplicationId;
+import co.cask.cdap.proto.id.ArtifactId;
+import co.cask.cdap.proto.id.DatasetId;
+import co.cask.cdap.proto.id.DatasetModuleId;
+import co.cask.cdap.proto.id.DatasetTypeId;
+import co.cask.cdap.proto.id.EntityId;
+import co.cask.cdap.proto.id.InstanceId;
+import co.cask.cdap.proto.id.NamespaceId;
+import co.cask.cdap.proto.id.ProgramId;
+import co.cask.cdap.proto.id.SecureKeyId;
+import co.cask.cdap.proto.id.StreamId;
+import co.cask.cdap.proto.security.Action;
+import co.cask.cdap.proto.security.Principal;
+import co.cask.cdap.proto.security.Privilege;
+import co.cask.cdap.proto.security.Role;
+import co.cask.cdap.security.spi.authorization.AbstractAuthorizer;
+import co.cask.cdap.security.spi.authorization.AuthorizationContext;
+import co.cask.cdap.security.spi.authorization.Authorizer;
+import co.cask.cdap.security.spi.authorization.UnauthorizedException;
+import com.google.common.base.Preconditions;
+import org.apache.hadoop.security.UserGroupInformation;
+import org.apache.ranger.audit.provider.MiscUtil;
+import org.apache.ranger.plugin.audit.RangerDefaultAuditHandler;
+import org.apache.ranger.plugin.policyengine.RangerAccessRequestImpl;
+import org.apache.ranger.plugin.policyengine.RangerAccessResourceImpl;
+import org.apache.ranger.plugin.policyengine.RangerAccessResult;
+import org.apache.ranger.plugin.service.RangerBasePlugin;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.net.InetAddress;
+import java.util.Date;
+import java.util.Properties;
+import java.util.Set;
+
+/**
+ * This class implements {@link Authorizer} from CDAP and is responsible for interacting with Ranger to enforce
+ * authorization.
+ */
+public class RangerAuthorizer extends AbstractAuthorizer {
+  private static final Logger LOG = LoggerFactory.getLogger(RangerAuthorizer.class);
+
+  // just string keys used to store entity in ranger. We don't want them to be derived from entity type or name since
+  // any changes to them on cdap side will make privileges incompatible.
+  private static final String KEY_INSTANCE = "instance";
+  private static final String KEY_NAMESPACE = "namespace";
+  private static final String KEY_ARTIFACT = "artifact";
+  private static final String KEY_APPLICATION = "application";
+  private static final String KEY_DATASET = "dataset";
+  private static final String KEY_STREAM = "stream";
+  private static final String KEY_PROGRAM = "program";
+  private static final String KEY_DATASET_MODULE = "dataset_module";
+  private static final String KEY_DATASET_TYPE = "dataset_type";
+  private static final String KEY_SECUREKEY = "securekey";
+
+  // using # as we don't allow it in entity names
+  private static final String RESOURCE_SEPARATOR = "#";
+
+  private static volatile RangerBasePlugin rangerPlugin = null;
+  private AuthorizationContext context;
+  // cdap instance name
+  private String instanceName;
+
+  @Override
+  public synchronized void initialize(AuthorizationContext context) throws Exception {
+    this.context = context;
+    Properties properties = context.getExtensionProperties();
+    instanceName = properties.containsKey(Constants.INSTANCE_NAME) ?
+      properties.getProperty(Constants.INSTANCE_NAME) : "cdap";
+    if (rangerPlugin == null) {
+      UserGroupInformation ugi = UserGroupInformation.getLoginUser();
+      Preconditions.checkNotNull(ugi, "Kerberos login information is not available. UserGroupInformation is null");
+      // set the login user as the user as whom cdap is running as this is needed for kerberos authentication
+      MiscUtil.setUGILoginUser(ugi, null);
+      LOG.debug("Initializing Ranger CDAP Plugin with UGI {}", ugi);
+
+      // the string name here should not be changed as this uniquely identifies the plugin in ranger. If it's
+      // changed it will require changing all the supporting xml file which is in this package.
+      rangerPlugin = new RangerBasePlugin("cdap", "cdap");
+    }
+    rangerPlugin.init();
+    RangerDefaultAuditHandler auditHandler = new RangerDefaultAuditHandler();
+    rangerPlugin.setResultProcessor(auditHandler);
+  }
+
+  @Override
+  public void enforce(EntityId entity, Principal principal, Action action) throws Exception {
+    LOG.debug("Enforce called on entity {}, principal {}, action {}", entity, principal, action);
+    if (rangerPlugin == null) {
+      throw new RuntimeException("CDAP Ranger Authorizer is not initialized.");
+    }
+
+    if (principal.getType() != Principal.PrincipalType.USER) {
+      throw new IllegalArgumentException(String.format("The principal type for current enforcement request is '%s'. " +
+                                                         "Authorization enforcement is only supported for '%s'.",
+                                                       principal.getType(), Principal.PrincipalType.USER));
+    }
+    String requestingUser = principal.getName();
+    String ip = InetAddress.getLocalHost().getHostName();
+    Set<String> userGroups = MiscUtil.getGroupsForRequestUser(requestingUser);
+
+    LOG.debug("Requesting user {}, ip {}, requesting user groups {}", requestingUser, ip, userGroups);
+
+    Date eventTime = new Date();
+    String accessType = toRangerAccessType(action);
+    RangerAccessRequestImpl rangerRequest = new RangerAccessRequestImpl();
+    rangerRequest.setUser(requestingUser);
+    rangerRequest.setUserGroups(userGroups);
+    rangerRequest.setClientIPAddress(ip);
+    rangerRequest.setAccessTime(eventTime);
+
+    RangerAccessResourceImpl rangerResource = new RangerAccessResourceImpl();
+    rangerRequest.setResource(rangerResource);
+    rangerRequest.setAccessType(accessType);
+    rangerRequest.setAction(accessType);
+    rangerRequest.setRequestData(entity.toString());
+
+    setAccessResource(entity, rangerResource);
+
+    boolean isAuthorized = false;
+
+    try {
+      RangerAccessResult result = rangerPlugin.isAccessAllowed(rangerRequest);
+      if (result == null) {
+        LOG.warn("Unauthorized: Ranger Plugin returned null for this authorization enforcement.");
+        isAuthorized = false;
+      } else {
+        isAuthorized = result.getIsAllowed();
+      }
+    } catch (Throwable t) {
+      LOG.warn("Error while calling isAccessAllowed(). request {}", rangerRequest, t);
+      throw t;
+    } finally {
+      LOG.trace("Ranger Request {}, authorization {}.", rangerRequest, (isAuthorized ? "successful" : "failed"));
+    }
+
+    if (!isAuthorized) {
+      throw new UnauthorizedException(principal, action, entity);
+    }
+  }
+
+  @Override
+  public void enforce(EntityId entityId, Principal principal, Set<Action> set) throws Exception {
+    LOG.debug("Enforce called on entity {}, principal {}, actions {}", entityId, principal, set);
+    //TODO: Investigate if its possible to make the enforce call with set of actions rather than one by one
+    for (Action action : set) {
+      enforce(entityId, principal, action);
+    }
+  }
+
+  @Override
+  public Set<? extends EntityId> isVisible(Set<? extends EntityId> set, Principal principal) throws Exception {
+    //TODO: Implement it.
+    return null;
+  }
+
+  @Override
+  public void grant(EntityId entity, Principal principal, java.util.Set<Action> actions) throws Exception {
+    LOG.warn("Grant operation not supported by Ranger for CDAP");
+  }
+
+  @Override
+  public void revoke(EntityId entity, Principal principal, java.util.Set<Action> actions) throws Exception {
+    LOG.warn("Revoke operation not supported by Ranger for CDAP");
+  }
+
+  @Override
+  public void revoke(EntityId entity) throws Exception {
+    LOG.warn("Revoke for entity operation not supported by Ranger for CDAP");
+  }
+
+  @Override
+  public void createRole(Role role) throws Exception {
+    LOG.warn("Create role operation not supported by Ranger for CDAP");
+
+  }
+
+  @Override
+  public void dropRole(Role role) throws Exception {
+    LOG.warn("Drop role operation not supported by Ranger for CDAP");
+
+  }
+
+  @Override
+  public void addRoleToPrincipal(Role role, Principal principal) throws Exception {
+    LOG.warn("Add role to principal operation not supported by Ranger for CDAP");
+
+  }
+
+  @Override
+  public void removeRoleFromPrincipal(Role role, Principal principal) throws Exception {
+    LOG.warn("Remove role from principal operation not supported by Ranger for CDAP");
+
+  }
+
+  @Override
+  public Set<Role> listRoles(Principal principal) throws Exception {
+    LOG.warn("List roles operation not supported by Ranger for CDAP");
+    return null;
+  }
+
+  @Override
+  public Set<Role> listAllRoles() throws Exception {
+    LOG.warn("List all roles operation not supported by Ranger for CDAP");
+    return null;
+  }
+
+  @Override
+  public Set<Privilege> listPrivileges(Principal principal) throws Exception {
+    LOG.warn("List privileges operation not supported by Ranger for CDAP");
+    return null;
+  }
+
+  private String toRangerAccessType(Action action) {
+    return action.toString().toLowerCase();
+  }
+
+  /**
+   * Sets the access resource appropriately depending on the given entityId
+   *
+   * @param entityId the entity which needs to be set to
+   * @param rangerAccessResource the {@link RangerAccessResourceImpl} to set the entity values to
+   */
+  private void setAccessResource(EntityId entityId, RangerAccessResourceImpl rangerAccessResource) {
+    EntityType entityType = entityId.getEntityType();
+    switch (entityType) {
+      case INSTANCE:
+        rangerAccessResource.setValue(KEY_INSTANCE, ((InstanceId) entityId).getInstance());
+        break;
+      case NAMESPACE:
+        setAccessResource(new InstanceId(instanceName), rangerAccessResource);
+        rangerAccessResource.setValue(KEY_NAMESPACE, ((NamespaceId) entityId).getNamespace());
+        break;
+      case ARTIFACT:
+        ArtifactId artifactId = (ArtifactId) entityId;
+        setAccessResource(artifactId.getParent(), rangerAccessResource);
+        rangerAccessResource.setValue(KEY_ARTIFACT, artifactId.getArtifact() + RESOURCE_SEPARATOR +
+          artifactId.getVersion());
+        break;
+      case APPLICATION:
+        ApplicationId applicationId = (ApplicationId) entityId;
+        setAccessResource(applicationId.getParent(), rangerAccessResource);
+        rangerAccessResource.setValue(KEY_APPLICATION, applicationId.getApplication());
+        break;
+      case DATASET:
+        DatasetId dataset = (DatasetId) entityId;
+        setAccessResource(dataset.getParent(), rangerAccessResource);
+        rangerAccessResource.setValue(KEY_DATASET, dataset.getDataset());
+        break;
+      case DATASET_MODULE:
+        DatasetModuleId datasetModuleId = (DatasetModuleId) entityId;
+        setAccessResource(datasetModuleId.getParent(), rangerAccessResource);
+        rangerAccessResource.setValue(KEY_DATASET_MODULE, datasetModuleId.getModule());
+        break;
+      case DATASET_TYPE:
+        DatasetTypeId datasetTypeId = (DatasetTypeId) entityId;
+        setAccessResource(datasetTypeId.getParent(), rangerAccessResource);
+        rangerAccessResource.setValue(KEY_DATASET_TYPE, datasetTypeId.getType());
+        break;
+      case STREAM:
+        StreamId streamId = (StreamId) entityId;
+        setAccessResource(streamId.getParent(), rangerAccessResource);
+        rangerAccessResource.setValue(KEY_STREAM, streamId.getStream());
+        break;
+      case PROGRAM:
+        ProgramId programId = (ProgramId) entityId;
+        setAccessResource(programId.getParent(), rangerAccessResource);
+        rangerAccessResource.setValue(KEY_PROGRAM, programId.getType() + RESOURCE_SEPARATOR + programId.getProgram());
+        break;
+      case SECUREKEY:
+        SecureKeyId secureKeyId = (SecureKeyId) entityId;
+        setAccessResource(secureKeyId.getParent(), rangerAccessResource);
+        rangerAccessResource.setValue(KEY_SECUREKEY, secureKeyId.getName());
+        break;
+      default:
+        throw new IllegalArgumentException(String.format("The entity %s is of unknown type %s", entityId, entityType));
+    }
+  }
+}
--- a/cdap-ranger/cdap-ranger-binding/src/main/resources/ranger-cdap-audit.xml
+++ b/cdap-ranger/cdap-ranger-binding/src/main/resources/ranger-cdap-audit.xml
@@ -1,0 +1,269 @@
+<?xml version="1.0"?>
+<!--
+  Licensed to the Apache Software Foundation (ASF) under one or more
+  contributor license agreements.  See the NOTICE file distributed with
+  this work for additional information regarding copyright ownership.
+  The ASF licenses this file to You under the Apache License, Version 2.0
+  (the "License"); you may not use this file except in compliance with
+  the License.  You may obtain a copy of the License at
+
+      http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+-->
+<?xml-stylesheet type="text/xsl" href="configuration.xsl"?>
+<configuration>
+  <property>
+    <name>xasecure.audit.is.enabled</name>
+    <value>true</value>
+  </property>
+
+  <!-- DB audit provider configuration -->
+  <property>
+    <name>xasecure.audit.db.is.enabled</name>
+    <value>false</value>
+  </property>
+
+  <property>
+    <name>xasecure.audit.db.is.async</name>
+    <value>true</value>
+  </property>
+
+  <property>
+    <name>xasecure.audit.db.async.max.queue.size</name>
+    <value>10240</value>
+  </property>
+
+  <property>
+    <name>xasecure.audit.db.async.max.flush.interval.ms</name>
+    <value>30000</value>
+  </property>
+
+  <property>
+    <name>xasecure.audit.db.batch.size</name>
+    <value>100</value>
+  </property>
+
+  <!--  Properties whose name begin with "xasecure.audit.jpa." are used to configure JPA -->
+  <property>
+    <name>xasecure.audit.jpa.javax.persistence.jdbc.url</name>
+    <value>jdbc:mysql://localhost:3306/ranger_audit</value>
+  </property>
+
+  <property>
+    <name>xasecure.audit.jpa.javax.persistence.jdbc.user</name>
+    <value>rangerlogger</value>
+  </property>
+
+  <property>
+    <name>xasecure.audit.jpa.javax.persistence.jdbc.password</name>
+    <value>none</value>
+  </property>
+
+  <property>
+    <name>xasecure.audit.jpa.javax.persistence.jdbc.driver</name>
+    <value>com.mysql.jdbc.Driver</value>
+  </property>
+
+  <property>
+    <name>xasecure.audit.credential.provider.file</name>
+    <value>jceks://file/etc/ranger/cdapdev/auditcred.jceks</value>
+  </property>
+
+
+  <!-- HDFS audit provider configuration -->
+  <property>
+    <name>xasecure.audit.hdfs.is.enabled</name>
+    <value>false</value>
+  </property>
+
+  <property>
+    <name>xasecure.audit.hdfs.is.async</name>
+    <value>true</value>
+  </property>
+
+  <property>
+    <name>xasecure.audit.hdfs.async.max.queue.size</name>
+    <value>1048576</value>
+  </property>
+
+  <property>
+    <name>xasecure.audit.hdfs.async.max.flush.interval.ms</name>
+    <value>30000</value>
+  </property>
+
+  <property>
+    <name>xasecure.audit.hdfs.config.encoding</name>
+    <value></value>
+  </property>
+
+  <property>
+    <name>xasecure.audit.hdfs.config.destination.directory</name>
+    <value>hdfs://NAMENODE_HOST:8020/ranger/audit/%app-type%/%time:yyyyMMdd%</value>
+  </property>
+
+  <property>
+    <name>xasecure.audit.hdfs.config.destination.file</name>
+    <value>%hostname%-audit.log</value>
+  </property>
+
+  <property>
+    <name>xasecure.audit.hdfs.config.destination.flush.interval.seconds</name>
+    <value>900</value>
+  </property>
+
+  <property>
+    <name>xasecure.audit.hdfs.config.destination.rollover.interval.seconds</name>
+    <value>86400</value>
+  </property>
+
+  <property>
+    <name>xasecure.audit.hdfs.config.destination.open.retry.interval.seconds</name>
+    <value>60</value>
+  </property>
+
+  <property>
+    <name>xasecure.audit.hdfs.config.local.buffer.directory</name>
+    <value>/var/log/cdap/audit</value>
+  </property>
+
+  <property>
+    <name>xasecure.audit.hdfs.config.local.buffer.file</name>
+    <value>%time:yyyyMMdd-HHmm.ss%.log</value>
+  </property>
+
+  <property>
+    <name>xasecure.audit.hdfs.config.local.buffer.file.buffer.size.bytes</name>
+    <value>8192</value>
+  </property>
+
+  <property>
+    <name>xasecure.audit.hdfs.config.local.buffer.flush.interval.seconds</name>
+    <value>60</value>
+  </property>
+
+  <property>
+    <name>xasecure.audit.hdfs.config.local.buffer.rollover.interval.seconds</name>
+    <value>600</value>
+  </property>
+
+  <property>
+    <name>xasecure.audit.hdfs.config.local.archive.directory</name>
+    <value>/var/log/cdap/audit/archive</value>
+  </property>
+
+  <property>
+    <name>xasecure.audit.hdfs.config.local.archive.max.file.count</name>
+    <value>10</value>
+  </property>
+
+  <!-- Audit to HDFS on Azure Datastore (WASB) requires v3 style settings.  Comment the above and uncomment only the
+  following to audit to Azure Blob Datastore via hdfs' WASB schema.
+
+  NOTE: If you specify one audit destination in v3 style then other destinations, if any, must also be specified in v3 style
+  ====
+
+  <property>
+    <name>xasecure.audit.destination.hdfs</name>
+    <value>enabled</value>
+  </property>
+
+  <property>
+    <name>xasecure.audit.destination.hdfs.dir</name>
+    <value>wasb://ranger-audit1@youraccount.blob.core.windows.net</value>
+  </property>
+
+  the following 3 correspond to the properties with similar name in core-site.xml, i.e.
+  - fs.azure.account.key.youraccount.blob.core.windows.net => xasecure.audit.destination.hdfs.config.fs.azure.account.key.youraccount.blob.core.windows.net and
+  - fs.azure.account.keyprovider.youraccount.blob.core.windows.net => xasecure.audit.destination.hdfs.config.fs.azure.account.keyprovider.youraccount.blob.core.windows.net,
+  - fs.azure.shellkeyprovider.script => xasecure.audit.destination.hdfs.config.fs.azure.shellkeyprovider.script,
+
+  <property>
+    <name>xasecure.audit.destination.hdfs.config.fs.azure.account.key.youraccount.blob.core.windows.net</name>
+    <value>YOUR ENCRYPTED ACCESS KEY</value>
+  </property>
+
+  <property>
+    <name>xasecure.audit.destination.hdfs.config.fs.azure.account.keyprovider.youraccount.blob.core.windows.net</name>
+    <value>org.apache.hadoop.fs.azure.ShellDecryptionKeyProvider</value>
+  </property>
+
+  <property>
+    <name>xasecure.audit.destination.hdfs.config.fs.azure.shellkeyprovider.script</name>
+    <value>/usr/lib/python2.7/dist-packages/hdinsight_common/decrypt.sh</value>
+  </property>
+  -->
+
+  <!-- Log4j audit provider configuration -->
+  <property>
+    <name>xasecure.audit.log4j.is.enabled</name>
+    <value>false</value>
+  </property>
+
+  <property>
+    <name>xasecure.audit.log4j.is.async</name>
+    <value>false</value>
+  </property>
+
+  <property>
+    <name>xasecure.audit.log4j.async.max.queue.size</name>
+    <value>10240</value>
+  </property>
+
+  <property>
+    <name>xasecure.audit.log4j.async.max.flush.interval.ms</name>
+    <value>30000</value>
+  </property>
+
+
+  <!-- Kafka audit provider configuration -->
+  <property>
+    <name>xasecure.audit.kafka.is.enabled</name>
+    <value>false</value>
+  </property>
+
+  <property>
+    <name>xasecure.audit.kafka.async.max.queue.size</name>
+    <value>1</value>
+  </property>
+
+  <property>
+    <name>xasecure.audit.kafka.async.max.flush.interval.ms</name>
+    <value>1000</value>
+  </property>
+
+  <property>
+    <name>xasecure.audit.kafka.broker_list</name>
+    <value>localhost:9092</value>
+  </property>
+
+  <property>
+    <name>xasecure.audit.kafka.topic_name</name>
+    <value>ranger_audits</value>
+  </property>
+
+  <!-- Ranger audit provider configuration -->
+  <property>
+    <name>xasecure.audit.solr.is.enabled</name>
+    <value>false</value>
+  </property>
+
+  <property>
+    <name>xasecure.audit.solr.async.max.queue.size</name>
+    <value>1</value>
+  </property>
+
+  <property>
+    <name>xasecure.audit.solr.async.max.flush.interval.ms</name>
+    <value>1000</value>
+  </property>
+
+  <property>
+    <name>xasecure.audit.solr.solr_url</name>
+    <value>http://localhost:6083/solr/ranger_audits</value>
+  </property>
+</configuration>
--- a/cdap-ranger/cdap-ranger-binding/src/main/resources/ranger-cdap-security.xml
+++ b/cdap-ranger/cdap-ranger-binding/src/main/resources/ranger-cdap-security.xml
@@ -1,0 +1,82 @@
+<?xml version="1.0"?>
+<!--
+  Licensed to the Apache Software Foundation (ASF) under one or more
+  contributor license agreements.  See the NOTICE file distributed with
+  this work for additional information regarding copyright ownership.
+  The ASF licenses this file to You under the Apache License, Version 2.0
+  (the "License"); you may not use this file except in compliance with
+  the License.  You may obtain a copy of the License at
+
+      http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+-->
+<?xml-stylesheet type="text/xsl" href="configuration.xsl"?>
+<configuration xmlns:xi="http://www.w3.org/2001/XInclude">
+  <property>
+    <name>ranger.plugin.cdap.service.name</name>
+    <value>cdapdev</value>
+    <description>
+      Name of the Ranger service containing policies for this CDAP instance
+    </description>
+  </property>
+
+  <property>
+    <name>ranger.plugin.cdap.policy.source.impl</name>
+    <value>org.apache.ranger.admin.client.RangerAdminRESTClient</value>
+    <description>
+      Class to retrieve policies from the source
+    </description>
+  </property>
+
+  <property>
+    <name>ranger.plugin.cdap.policy.rest.url</name>
+    <value>http://host:port</value>
+    <description>
+      URL to Ranger Admin
+    </description>
+  </property>
+
+  <property>
+    <name>ranger.plugin.cdap.policy.rest.ssl.config.file</name>
+    <value>/etc/cdap/conf/ranger-policymgr-ssl.xml</value>
+    <description>
+      Path to the file containing SSL details to contact Ranger Admin
+    </description>
+  </property>
+
+  <property>
+    <name>ranger.plugin.cdap.policy.pollIntervalMs</name>
+    <value>60000</value>
+    <description>
+      How often to poll for changes in policies?
+    </description>
+  </property>
+  <property>
+    <name>ranger.plugin.cdap.policy.cache.dir</name>
+    <value>/etc/ranger/cdapdev/policycache</value>
+    <description>
+      Directory where Ranger policies are cached after successful retrieval from the source
+    </description>
+  </property>
+
+  <property>
+    <name>ranger.plugin.cdap.policy.rest.client.connection.timeoutMs</name>
+    <value>120000</value>
+    <description>
+      RangerRestClient Connection Timeout in Milli Seconds
+    </description>
+  </property>
+
+  <property>
+    <name>ranger.plugin.cdap.policy.rest.client.read.timeoutMs</name>
+    <value>30000</value>
+    <description>
+      RangerRestClient read Timeout in Milli Seconds
+    </description>
+  </property>
+</configuration>
--- a/cdap-ranger/cdap-ranger-binding/src/main/resources/ranger-policymgr-ssl.xml
+++ b/cdap-ranger/cdap-ranger-binding/src/main/resources/ranger-policymgr-ssl.xml
@@ -1,0 +1,63 @@
+<?xml version="1.0"?>
+<!--
+  Licensed to the Apache Software Foundation (ASF) under one or more
+  contributor license agreements.  See the NOTICE file distributed with
+  this work for additional information regarding copyright ownership.
+  The ASF licenses this file to You under the Apache License, Version 2.0
+  (the "License"); you may not use this file except in compliance with
+  the License.  You may obtain a copy of the License at
+
+      http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+-->
+<?xml-stylesheet type="text/xsl" href="configuration.xsl"?>
+<configuration xmlns:xi="http://www.w3.org/2001/XInclude">
+	<!--  The following properties are used for 2-way SSL client server validation -->
+	<property>
+		<name>xasecure.policymgr.clientssl.keystore</name>
+		<value>cdap-clientcert.jks</value>
+		<description> 
+			Java Keystore files 
+		</description>
+	</property>
+	<property>
+		<name>xasecure.policymgr.clientssl.keystore.password</name>
+		<value>none</value>
+		<description> 
+			password for keystore 
+		</description>
+	</property>
+	<property>
+		<name>xasecure.policymgr.clientssl.truststore</name>
+		<value>cacerts-xasecure.jks</value>
+		<description> 
+			java truststore file
+		</description>
+	</property>
+	<property>
+		<name>xasecure.policymgr.clientssl.truststore.password</name>
+		<value>none</value>
+		<description> 
+			java  truststore password
+		</description>
+	</property>
+    <property>
+		<name>xasecure.policymgr.clientssl.keystore.credential.file</name>
+		<value>jceks://file/tmp/keystore-kafkadev-ssl.jceks</value>
+		<description> 
+			java  keystore credential file
+		</description>
+	</property>
+	<property>
+		<name>xasecure.policymgr.clientssl.truststore.credential.file</name>
+		<value>jceks://file/tmp/truststore-kafkadev-ssl.jceks</value>
+		<description> 
+			java  truststore credential file
+		</description>
+	</property>
+</configuration>
--- a/cdap-ranger/cdap-ranger-lookup/pom.xml
+++ b/cdap-ranger/cdap-ranger-lookup/pom.xml
@@ -1,0 +1,71 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  Copyright © 2017 Cask Data, Inc.
+
+  Licensed under the Apache License, Version 2.0 (the "License"); you may not
+  use this file except in compliance with the License. You may obtain a copy of
+  the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+  WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+  License for the specific language governing permissions and limitations under
+  the License.
+  -->
+
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <parent>
+    <artifactId>cdap-ranger</artifactId>
+    <groupId>co.cask.cdap</groupId>
+    <version>0.1.0-SNAPSHOT</version>
+  </parent>
+  <modelVersion>4.0.0</modelVersion>
+
+  <artifactId>cdap-ranger-lookup</artifactId>
+
+  <name>CDAP Ranger Lookup</name>
+  <description>Defines the CDAP Data Model in Apache Ranger and provided looking up CDAP resources.</description>
+
+  <dependencies>
+    <dependency>
+      <groupId>org.apache.ranger</groupId>
+      <artifactId>ranger-plugins-common</artifactId>
+      <scope>provided</scope>
+    </dependency>
+    <dependency>
+      <groupId>co.cask.cdap</groupId>
+      <artifactId>cdap-client</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>co.cask.cdap</groupId>
+      <artifactId>cdap-cli</artifactId>
+    </dependency>
+  </dependencies>
+
+  <build>
+    <plugins>
+      <plugin>
+        <!-- used to build a flat uber jar which can be placed in apache ranger classpath -->
+        <artifactId>maven-assembly-plugin</artifactId>
+        <configuration>
+          <descriptorRefs>
+            <descriptorRef>jar-with-dependencies</descriptorRef>
+          </descriptorRefs>
+        </configuration>
+        <executions>
+          <execution>
+            <id>make-assembly</id> <!-- this is used for inheritance merges -->
+            <phase>package</phase> <!-- bind to the packaging phase -->
+            <goals>
+              <goal>single</goal>
+            </goals>
+          </execution>
+        </executions>
+      </plugin>
+    </plugins>
+  </build>
+</project>
--- a/cdap-ranger/cdap-ranger-lookup/src/main/java/co/cask/cdap/security/authorization/ranger/lookup/RangerLookupService.java
+++ b/cdap-ranger/cdap-ranger-lookup/src/main/java/co/cask/cdap/security/authorization/ranger/lookup/RangerLookupService.java
@@ -1,0 +1,92 @@
+/*
+ * Copyright © 2017 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package co.cask.cdap.security.authorization.ranger.lookup;
+
+import co.cask.cdap.security.authorization.ranger.lookup.client.CDAPConnectionManager;
+import co.cask.cdap.security.authorization.ranger.lookup.client.CDAPRangerLookupClient;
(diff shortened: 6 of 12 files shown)


cdap-ranger/cdap-ranger-binding/src/main/java/co/cask/cdap/security/authorization/ranger/binding/RangerAuthorizer.java
```diff
@@ -0,0 +1,288 @@
...
+      properties.getProperty(Constants.INSTANCE_NAME) : "cdap";
+    if (rangerPlugin == null) {
+      try {
+        UserGroupInformation ugi = UserGroupInformation.getLoginUser();
+        if (ugi != null) {
+          MiscUtil.setUGILoginUser(ugi, null);
+        }
+        LOG.debug("Initializing Ranger CDAP Plugin with UGI {}", ugi);
+      } catch (Throwable t) {
+        LOG.error("Error getting principal.", t);
```
anwar6953: shouldn't this be fatal? (i.e. don't just log it, but throw the exception)

cdap-ranger/cdap-ranger-binding/src/main/java/co/cask/cdap/security/authorization/ranger/binding/RangerAuthorizer.java
```diff
@@ -0,0 +1,288 @@
...
+    rangerPlugin.setResultProcessor(auditHandler);
+  }
+
+
+  @Override
+  public void enforce(EntityId entity, Principal principal, Action action) throws Exception {
+    LOG.info("=====> enforce({}, {}, {})", entity, principal, action);
+    LOG.info("Enforce called on entity {}, principal {}, action {}", entity, principal, action);
+    if (rangerPlugin == null) {
+      LOG.warn("CDAP Ranger Authorizer is not initialized");
```
anwar6953: why LOG AND throw?
Simply throwing the exception should be sufficient, right?

cdap-ranger/cdap-ranger-binding/src/main/java/co/cask/cdap/security/authorization/ranger/binding/RangerAuthorizer.java
```diff
@@ -0,0 +1,288 @@
...
+    rangerRequest.setAccessType(accessType);
+    rangerRequest.setAction(accessType);
+    rangerRequest.setRequestData(entity.toString());
+
+    setAccessResource(entity, rangerResource);
+
+    boolean isAuthorized = false;
+
+    try {
+      RangerAccessResult result = rangerPlugin.isAccessAllowed(rangerRequest);
```
anwar6953, line 142: can it occasionally return null?

cdap-ranger/cdap-ranger-lookup/src/main/java/co/cask/cdap/security/authorization/ranger/lookup/RangerLookupService.java
```diff
@@ -0,0 +1,92 @@
...
+      LOG.debug("==> RangerLookupService.validateConfig(" + serviceName + ")");
+    }
+
+    // HashMap since validateConfig api returns a HashMap.
+    HashMap<String, Object> response = new HashMap<>();
+    if (configs != null) {
+      try {
+        response.putAll(CDAPConnectionMgr.testConnection(serviceName, configs));
+      } catch (HadoopException e) {
+        LOG.error("<== RangerLookupService.validateConfig Error:" + e);
```
anwar6953: why log AND throw?

rohitsinha54: Oops.. most of these logs where for debugging. I think the commit got lost in my rebase somewhere . Will pick it

rohitsinha54: The log since this goes in ranger log and its important to log which plugin exactly threw it validateConfig is an abstract method which all plugins must implement. 
The exception is thrown again so that ranger can handle it and failing the connection test call and showing the right message in the UI.

cdap-ranger/cdap-ranger-binding/src/main/java/co/cask/cdap/security/authorization/ranger/binding/RangerAuthorizer.java
rohitsinha54: Good point. Although I think we need to decide as what's the right behavior for cdap when it fails to instantiate authorization plugin. Should cdap startup itself fail or should it start and then all the authorization will fail. First has the advantage that of failing early on although it limits the user in being able to start cdap successfully. The second will make cdap start although the most of the operations will fail since all the authorization will fail. Since enforce check for initialization the error the user will see will point to the cause.
@poornachandra @anwar6953 : Thoughts ?

cdap-ranger/cdap-ranger-binding/src/main/java/co/cask/cdap/security/authorization/ranger/binding/RangerAuthorizer.java
rohitsinha54: The underlying implementation does return null.

cdap-ranger/cdap-ranger-lookup/src/main/java/co/cask/cdap/security/authorization/ranger/lookup/RangerLookupService.java
rohitsinha54: The log since this goes in ranger log and its important to log which plugin exactly threw it validateConfig is an abstract method which all plugins must implement.
The exception is thrown again so that ranger can handle it and failing the connection test call and showing the right message in the UI.

rohitsinha54 (commented):
Some of the log statements was for debugging. I think I lost the commit which removed it in rebase.

cdap-ranger/cdap-ranger-binding/src/main/java/co/cask/cdap/security/authorization/ranger/binding/RangerAuthorizer.java
```diff
@@ -0,0 +1,306 @@
...
+    instanceName = properties.containsKey(Constants.INSTANCE_NAME) ?
+      properties.getProperty(Constants.INSTANCE_NAME) : "cdap";
+    if (rangerPlugin == null) {
+      try {
+        UserGroupInformation ugi = UserGroupInformation.getLoginUser();
+        if (ugi != null) {
+          // set the login user as the user as whom cdap is running as this is needed for kerberos authentication
+          MiscUtil.setUGILoginUser(ugi, null);
+        }
+        LOG.debug("Initializing Ranger CDAP Plugin with UGI {}", ugi);
```
yaojiefeng: Should we throw exception or at least log some messages if the login user is null since that means something bad happens?

cdap-ranger/cdap-ranger-binding/src/main/java/co/cask/cdap/security/authorization/ranger/binding/RangerAuthorizer.java
```diff
@@ -0,0 +1,306 @@
...
+      } catch (Throwable t) {
+        LOG.error("Error getting principal.", t);
+      }
+      // the string name here should not be changed as this uniquely identifies the service in ranger. If it's
+      // changed it will require changing all the supporting xml file which is in this package.
+      rangerPlugin = new RangerBasePlugin("cdap", "cdap");
```
yaojiefeng: Maybe we can file a JIRA to make this configurable in the future?

cdap-ranger/cdap-ranger-binding/src/main/java/co/cask/cdap/security/authorization/ranger/binding/RangerAuthorizer.java
```diff
@@ -0,0 +1,306 @@
...
+    }
+    rangerPlugin.init();
+    RangerDefaultAuditHandler auditHandler = new RangerDefaultAuditHandler();
+    rangerPlugin.setResultProcessor(auditHandler);
+  }
+
+
```
yaojiefeng: extra new line

cdap-ranger/cdap-ranger-binding/src/main/java/co/cask/cdap/security/authorization/ranger/binding/RangerAuthorizer.java
```diff
@@ -0,0 +1,306 @@
...
+      LOG.warn("Error while calling isAccessAllowed(). request {}", rangerRequest, t);
+      throw t;
+    } finally {
+      LOG.debug("Ranger Request {}, authorization {}.", rangerRequest, (isAuthorized ? "successful" : "failed"));
+    }
+
+    if (!isAuthorized) {
+      // explicitly log here since we have seen cases where UnauthorizedException doesn't get logged by cdap and causes
+      // debugging nightmare.
+      LOG.warn("Unauthorized: Principal {} is unauthorized to perform action {} on entity {}.",
```
yaojiefeng: Should we reply on CDAP side to actually log the unauthorized exception? Since we log the result of the enforce in debug level, I don't think it is very helpful to log warn here since we can enable debug in this class to see the result of the enforce

cdap-ranger/cdap-ranger-binding/src/main/java/co/cask/cdap/security/authorization/ranger/binding/RangerAuthorizer.java
```diff
@@ -0,0 +1,306 @@
...
+    rangerRequest.setAction(accessType);
+    rangerRequest.setRequestData(entity.toString());
+
+    setAccessResource(entity, rangerResource);
+
+    boolean isAuthorized = false;
+
+    try {
+      RangerAccessResult result = rangerPlugin.isAccessAllowed(rangerRequest);
+      if (result == null) {
```
yaojiefeng, line 143: I see in the implementation of `isAccessAllowed ` method, it returns null if the policyEngine is null. May I ask what does that mean?

cdap-ranger/cdap-ranger-binding/src/main/java/co/cask/cdap/security/authorization/ranger/binding/RangerAuthorizer.java
```diff
@@ -0,0 +1,306 @@
...
+   * @param rangerAccessResource the {@link RangerAccessResourceImpl} to set the entity values to
+   */
+  private void setAccessResource(EntityId entityId, RangerAccessResourceImpl rangerAccessResource) {
+    EntityType entityType = entityId.getEntityType();
+    switch (entityType) {
+      case INSTANCE:
+        rangerAccessResource.setValue(KEY_INSTANCE, ((InstanceId) entityId).getInstance());
+        break;
+      case NAMESPACE:
+        rangerAccessResource.setValue(KEY_INSTANCE, instanceName);
```
yaojiefeng: Should we put instanceid as the value here since `InstanceId` is not a child of `NamespaceId`?

cdap-ranger/cdap-ranger-binding/src/main/resources/ranger-cdap-audit.xml
```diff
@@ -0,0 +1,269 @@
...
+
+  <property>
+    <name>xasecure.audit.db.batch.size</name>
+    <value>100</value>
+  </property>
+
+  <!--  Properties whose name begin with "xasecure.audit.jpa." are used to configure JPA -->
+  <property>
+    <name>xasecure.audit.jpa.javax.persistence.jdbc.url</name>
+    <value>jdbc:mysql://localhost:3306/ranger_audit</value>
```
yaojiefeng, line 54: Can you explain a bit what this file is used for? Is it possible that the user `rangerlogger` does not exist or does not have privilege for mysql?

cdap-ranger/cdap-ranger-binding/src/main/resources/ranger-cdap-security.xml
```diff
@@ -0,0 +1,82 @@
...
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+-->
+<?xml-stylesheet type="text/xsl" href="configuration.xsl"?>
+<configuration xmlns:xi="http://www.w3.org/2001/XInclude">
+  <property>
+    <name>ranger.plugin.cdap.service.name</name>
+    <value>cdapdev</value>
```
yaojiefeng, line 22: Is this the name for the ranger plugin here: https://github.com/caskdata/cdap-security-extn/blob/a84aee903bf12a9e82d845453d7fec4e28f94a03/cdap-ranger/cdap-ranger-binding/src/main/java/co/cask/cdap/security/authorization/ranger/binding/RangerAuthorizer.java#L101, should it be cdap or cdapdev?

cdap-ranger/cdap-ranger-binding/src/main/resources/ranger-cdap-security.xml
```diff
@@ -0,0 +1,82 @@
...
+  <property>
+    <name>ranger.plugin.cdap.policy.pollIntervalMs</name>
+    <value>60000</value>
+    <description>
+      How often to poll for changes in policies?
+    </description>
+  </property>
+  <property>
+    <name>ranger.plugin.cdap.policy.cache.dir</name>
+    <value>/etc/ranger/cdapdev/policycache</value>
```
yaojiefeng, line 61: Do we know what is the timeout for this cache?

cdap-ranger/cdap-ranger-binding/src/main/resources/ranger-policymgr-ssl.xml
```diff
@@ -0,0 +1,63 @@
...
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+-->
+<?xml-stylesheet type="text/xsl" href="configuration.xsl"?>
+<configuration xmlns:xi="http://www.w3.org/2001/XInclude">
+	<!--  The following properties are used for 2-way SSL client server validation -->
```
yaojiefeng, line 20: The indentation seems off for this file

yaojiefeng: Also this file contains the config for some password, should it be configurable by user?

cdap-ranger/cdap-ranger-lookup/src/main/java/co/cask/cdap/security/authorization/ranger/lookup/RangerLookupService.java
```diff
@@ -0,0 +1,92 @@
...
+  }
+
+  @Override
+  public void init(RangerServiceDef serviceDef, RangerService service) {
+    super.init(serviceDef, service);
+  }
+
+  @Override
+  public HashMap<String, Object> validateConfig() throws Exception {
+    if (LOG.isDebugEnabled()) {
```
yaojiefeng, line 52: Do we need this check here? Can we directly do `LOG.debug(...)`? Since the default log level for us is INFO, I dont think it will log anything if we have this check.

cdap-ranger/cdap-ranger-lookup/src/main/java/co/cask/cdap/security/authorization/ranger/lookup/RangerLookupService.java
```diff
@@ -0,0 +1,92 @@
...
+    HashMap<String, Object> response = new HashMap<>();
+    if (configs != null) {
+      try {
+        response.putAll(CDAPConnectionMgr.testConnection(serviceName, configs));
+      } catch (HadoopException e) {
+        LOG.error("<== RangerLookupService.validateConfig Error:" + e);
+        throw e;
+      }
+    }
+    if (LOG.isDebugEnabled()) {
```
yaojiefeng, line 66: Same here. Can we directly do `LOG.debug(...)`

cdap-ranger/cdap-ranger-lookup/src/main/java/co/cask/cdap/security/authorization/ranger/lookup/RangerLookupService.java
```diff
@@ -0,0 +1,92 @@
...
+  @Override
+  public List<String> lookupResource(ResourceLookupContext context) throws Exception {
+    List<String> ret = null;
+    if (LOG.isDebugEnabled()) {
+      LOG.debug("==> RangerLookupService.lookupResource Context: (" + context + ")");
+    }
+    if (context != null) {
+      try {
+        CDAPClient client = CDAPConnectionMgr.getCDAPClient(serviceName, configs);
+        // TODO do resource lookup here
```
yaojiefeng: Should we file a JIRA and put it here?

cdap-ranger/cdap-ranger-lookup/src/main/java/co/cask/cdap/security/authorization/ranger/lookup/client/CDAPClient.java
```diff
@@ -0,0 +1,226 @@
...
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+
+/**
+ * CDAP Clients which to be used by ranger for listing cdap resources.
+ * Note: The log statements in this class are written in a formatted way to match the logging format
+ * used by other services in Apache Ranger since these log statements end up in ranger admin log file we
+ * want them to be in same format.
+ */
+public class CDAPClient {
```
yaojiefeng: Is `CDAPRangerLookUpClient` a better name? In future, we may have client for Sentry for grant/revoke. It is good to distinguish them.

cdap-ranger/cdap-ranger-lookup/src/main/java/co/cask/cdap/security/authorization/ranger/lookup/client/CDAPConnectionMgr.java
```diff
@@ -0,0 +1,58 @@
...
+package co.cask.cdap.security.authorization.ranger.lookup.client;
+
+import com.google.common.base.Strings;
+
+import java.util.Map;
+
+/**
+ * CDAP Client and Connection Manager.
+ */
+public class CDAPConnectionMgr {
```
yaojiefeng: Is `CDAPRangerConnectionManager` a better name? I think we should use full name as class name.

cdap-ranger/cdap-ranger-lookup/src/main/java/co/cask/cdap/security/authorization/ranger/lookup/client/CDAPClient.java
```diff
@@ -0,0 +1,226 @@
...
+   * @return Map<String, Object> Connection test response
+   */
+  Map<String, Object> testConnection() {
+    if (LOG.isDebugEnabled()) {
+      LOG.debug("==> CDAPClient testConnection()");
+    }
+    HashMap<String, Object> responseData = new HashMap<>();
+    List<String> testResult;
+    try {
+      testResult = getNamespaces(null);
```
yaojiefeng: We can pass in an empty list here so that in the method implementation we do not need to check for null

cdap-ranger/cdap-ranger-lookup/src/main/java/co/cask/cdap/security/authorization/ranger/lookup/client/CDAPClient.java
```diff
@@ -0,0 +1,226 @@
...
+  private static final Log LOG = LogFactory.getLog(CDAPClient.class);
+  private static final long SERVICE_TIMEOUT_SECONDS = TimeUnit.MINUTES.toSeconds(30);
+
+  private AccessToken accessToken;
+  private final String serviceName;
+  private final String instanceURL;
+  private final String username;
+  private final String password;
+  private final NamespaceClient nsClient;
+  private static final String ERR_MSG = " You can still save the repository and start creating "
```
yaojiefeng: nit: static variable should go up

cdap-ranger/cdap-ranger-lookup/src/main/java/co/cask/cdap/security/authorization/ranger/lookup/client/CDAPClient.java
```diff
@@ -0,0 +1,226 @@
...
+
+    List<String> namespaces = new ArrayList<>();
+    if (nsClient != null) {
+      for (NamespaceMeta namespaceMeta : nsClient.list()) {
+        String name = namespaceMeta.getName();
+        if (nsList == null || !nsList.contains(name)) {
+          namespaces.add(name);
+        }
+      }
+    } else {
```
yaojiefeng: When can nsClient be null?

cdap-ranger/cdap-ranger-lookup/src/main/java/co/cask/cdap/security/authorization/ranger/lookup/client/CDAPClient.java
```diff
@@ -0,0 +1,226 @@
...
+   */
+  Map<String, Object> testConnection() {
+    if (LOG.isDebugEnabled()) {
+      LOG.debug("==> CDAPClient testConnection()");
+    }
+    HashMap<String, Object> responseData = new HashMap<>();
+    List<String> testResult;
+    try {
+      testResult = getNamespaces(null);
+      if (testResult != null && testResult.size() >= 0) {
```
yaojiefeng: the testResult will not null here since you initialize it as an empty list. Also, should the condition be `testResult.size() > 0`?

cdap-ranger/cdap-ranger-lookup/src/main/java/co/cask/cdap/security/authorization/ranger/lookup/client/CDAPClient.java
```diff
@@ -0,0 +1,226 @@
...
+        String successMsg = "Connection Test Successful";
+        BaseClient.generateResponseDataMap(true, successMsg, successMsg,
+                                           null, null, responseData);
+      } else {
```
yaojiefeng: So the definition of testConnection() is the user can at least list one namespace. I am not sure if this is the correct way, since user can have privileges on just some entities other than namespace. Maybe we should consider the connection is successful if list() does not throw any exception?

cdap-ranger/cdap-ranger-lookup/src/main/java/co/cask/cdap/security/authorization/ranger/lookup/client/CDAPClient.java
```diff
@@ -0,0 +1,226 @@
...
+      BaseClient.generateResponseDataMap(false, failureMsg,
+                                         failureMsg + ERR_MSG, null, null, responseData);
+    }
+    if (LOG.isDebugEnabled()) {
+      LOG.debug("<== CDAPClient testConnection()");
+    }
+    return responseData;
+  }
+
+  private List<String> getNamespaces(List<String> nsList) throws Exception {
```
yaojiefeng: What is the usage of this method? The only place it is used is to get the namespaces for test connection. If that is the case why do we need to pass a list here, and we can just return `nsClient.list()`?

cdap-ranger/cdap-ranger-lookup/src/main/java/co/cask/cdap/security/authorization/ranger/lookup/client/CDAPConnectionMgr.java
```diff
@@ -0,0 +1,58 @@
...
+
+  private static final String INSTANCE_URL = "cdap.instance.url";
+  private static final String USERNAME = "cdap.username";
+  private static final String PASSWORD = "cdap.password";
+
+  /**
+   * Returns a CDAP Client containing namespace, stream, dataset etc clients
```
yaojiefeng: nit: need a new line after this

cdap-ranger/cdap-ranger-lookup/src/main/java/co/cask/cdap/security/authorization/ranger/lookup/client/CDAPConnectionMgr.java
```diff
@@ -0,0 +1,58 @@
...
+    String password = configs.get(PASSWORD);
+    if (!(Strings.isNullOrEmpty(instanceURL) || Strings.isNullOrEmpty(username) || Strings.isNullOrEmpty(password))) {
+      return new CDAPClient(serviceName, instanceURL, username, password);
+    }
+    throw new IllegalArgumentException("Required properties are not set for "
+                          + serviceName + ". CDAP instance url with port, username and password must be provided.");
+  }
+
+  /**
+   * Tests that connection to CDAP instance can be made with the given config
```
yaojiefeng: nit: need a new line after this

yaojiefeng (commented):
Took a pass, have some comments

cdap-ranger/cdap-ranger-binding/src/main/java/co/cask/cdap/security/authorization/ranger/binding/RangerAuthorizer.java
rohitsinha54: I don't think it needs to configurable. This will require a change only if we start calling cdap something else

cdap-ranger/cdap-ranger-binding/src/main/java/co/cask/cdap/security/authorization/ranger/binding/RangerAuthorizer.java
rohitsinha54: It means that authorization enforcement didn't happen so ranger cannot tell you whether the access is allowed or not. Now it's up to the called how to handle it whether throw exception deny authorization.

cdap-ranger/cdap-ranger-binding/src/main/resources/ranger-cdap-audit.xml
rohitsinha54: I don't understand this comment. This is ranger side configuration. We don't need to handle anything here. User is responsible for configuring ranger correctly.

cdap-ranger/cdap-ranger-binding/src/main/resources/ranger-cdap-security.xml
rohitsinha54: It is the name of the this service it can be cdapdev or cdap. Although I see your confusion. I updated the comment in the other place to call it plugin name.

cdap-ranger/cdap-ranger-binding/src/main/resources/ranger-cdap-security.xml
rohitsinha54: ya its few lines above. currently its set  to 60000 ms

cdap-ranger/cdap-ranger-binding/src/main/resources/ranger-policymgr-ssl.xml
rohitsinha54: This is sample xml file for ssl it needs to be configured by the user if ssl is enabled. We need to look into how to expose it though since its get packaged in jar. That we will do at the end.

cdap-ranger/cdap-ranger-lookup/src/main/java/co/cask/cdap/security/authorization/ranger/lookup/RangerLookupService.java
rohitsinha54: This runs inside ranger their logging is different than ours. See the comment on the class about  log.

cdap-ranger/cdap-ranger-lookup/src/main/java/co/cask/cdap/security/authorization/ranger/lookup/RangerLookupService.java
rohitsinha54: No its coming in another PR. Its a crucial part of ranger integration and is tracked by this same jira this pr belongs to

cdap-ranger/cdap-ranger-lookup/src/main/java/co/cask/cdap/security/authorization/ranger/lookup/client/CDAPClient.java
rohitsinha54: 👍

cdap-ranger/cdap-ranger-lookup/src/main/java/co/cask/cdap/security/authorization/ranger/lookup/client/CDAPClient.java
rohitsinha54: In this case we need null. This look is going to involve for partial lookup for completion in UI and it will be needed there.

cdap-ranger/cdap-ranger-lookup/src/main/java/co/cask/cdap/security/authorization/ranger/lookup/client/CDAPClient.java
rohitsinha54: Removed the null check 👍 
It checking for equal to 0 because its possible that we see an empty list which will be the case when authentication is successful but the user has no privilege on any ns.

cdap-ranger/cdap-ranger-lookup/src/main/java/co/cask/cdap/security/authorization/ranger/lookup/client/CDAPClient.java
rohitsinha54: See above comment. That is why check if >= 0

cdap-ranger/cdap-ranger-lookup/src/main/java/co/cask/cdap/security/authorization/ranger/lookup/client/CDAPClient.java
rohitsinha54: Its going to be used in ns lookup in next PR

cdap-ranger/cdap-ranger-lookup/src/main/java/co/cask/cdap/security/authorization/ranger/lookup/client/CDAPClient.java
rohitsinha54: When the initialization failed.

cdap-ranger/cdap-ranger-lookup/src/main/java/co/cask/cdap/security/authorization/ranger/lookup/client/CDAPConnectionMgr.java
rohitsinha54: Ranger names all their plugin manager class as Mgr. I was keeping it consistent as this runs inside in ranger. But you are right it looks ugly. Will change it.

cdap-ranger/cdap-ranger-binding/src/main/resources/ranger-policymgr-ssl.xml
yaojiefeng: I see. Should we report a JIRA about this?

cdap-ranger/cdap-ranger-binding/src/main/resources/ranger-cdap-audit.xml
yaojiefeng: I see.

cdap-ranger/cdap-ranger-lookup/src/main/java/co/cask/cdap/security/authorization/ranger/lookup/RangerLookupService.java
yaojiefeng: I see. So it is user's responsibility to configure the logger for ranger, right?

cdap-ranger/cdap-ranger-lookup/src/main/java/co/cask/cdap/security/authorization/ranger/lookup/client/CDAPClient.java
yaojiefeng: Yeah, I get the point that the returned size can be 0. But the else statement is useless and the message it is reporting is to indicate the user should have access to at least one namespace there. Should remove the else

cdap-ranger/cdap-ranger-lookup/src/main/java/co/cask/cdap/security/authorization/ranger/lookup/client/CDAPClient.java
yaojiefeng: I still dont understand why we need pass null. In the method implementation, we are excluding all the namespaces in this list. If we pass empty list, the functionality seems same to me.

cdap-ranger/cdap-ranger-lookup/src/main/java/co/cask/cdap/security/authorization/ranger/lookup/client/CDAPRangerLookupClient.java
```diff
@@ -0,0 +1,227 @@
...
+   * @return Map<String, Object> Connection test response
+   */
+  Map<String, Object> testConnection() {
+    if (LOG.isDebugEnabled()) {
+      LOG.debug("==> CDAPRangerLookupClient testConnection()");
+    }
+    HashMap<String, Object> responseData = new HashMap<>();
+    List<String> testResult;
+    try {
+      testResult = getNamespaces(null);
```
yaojiefeng: I still dont understand why we need pass null. In the method implementation, we are excluding all the namespaces in this list. If we pass empty list, the functionality seems same to me.

cdap-ranger/cdap-ranger-lookup/src/main/java/co/cask/cdap/security/authorization/ranger/lookup/client/CDAPRangerLookupClient.java
```diff
@@ -0,0 +1,227 @@
...
+      if (testResult.size() >= 0) {
+        String successMsg = "Connection Test Successful";
+        BaseClient.generateResponseDataMap(true, successMsg, successMsg,
+                                           null, null, responseData);
+      } else {
```
yaojiefeng: Yeah, I get the point that the returned size can be 0. But the else statement is useless and the message it is reporting is to indicate the user should have access to at least one namespace there. Should remove the else

rohitsinha54: Oops  else was from earlier code which didnt handle 0. Nice catch. Removed it.

cdap-ranger/cdap-ranger-lookup/src/main/java/co/cask/cdap/security/authorization/ranger/lookup/client/CDAPRangerLookupClient.java
rohitsinha54: So there is two places where this will be used. First what you see right now testConnection and later in the lookup. The lookup will pass a empty or partial list of namespaces which will get modified. The testConnection passes null because for test connection we are not interested in looking what is in the list and hence there is no need to call contains. Since its or operation if the first statement implied to true the second is not evaluated hence we don't check contains (since it's a list and the contains time complexity is O(n) here)

cdap-ranger/cdap-ranger-binding/src/main/java/co/cask/cdap/security/authorization/ranger/binding/RangerAuthorizer.java
```diff
@@ -0,0 +1,301 @@
...
+        Preconditions.checkNotNull(ugi, "UserGroupInformation is null");
+        // set the login user as the user as whom cdap is running as this is needed for kerberos authentication
+        MiscUtil.setUGILoginUser(ugi, null);
+        LOG.debug("Initializing Ranger CDAP Plugin with UGI {}", ugi);
+      } catch (Throwable t) {
+        LOG.error("Error getting principal.", t);
+      }
+      // the string name here should not be changed as this uniquely identifies the plugin in ranger. If it's
+      // changed it will require changing all the supporting xml file which is in this package.
+      rangerPlugin = new RangerBasePlugin("cdap", "cdap");
```
yaojiefeng: Can we add more comments on which xml actually has this value? In `range-cdap-audit.xml`, I see the property `ranger.plugin.cdap.service.name` named as `cdapdev`, which really confuses me.

yaojiefeng (commented):
Just one comment, rest LGTM

cdap-ranger/cdap-ranger-binding/src/main/java/co/cask/cdap/security/authorization/ranger/binding/RangerAuthorizer.java
rohitsinha54: those are different. The comment says what needs to be changed. 
> it will require changing all the supporting xml file which is in this package.

cdap-ranger/cdap-ranger-binding/src/main/java/co/cask/cdap/security/authorization/ranger/binding/RangerAuthorizer.java
```diff
@@ -0,0 +1,301 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.net.InetAddress;
+import java.util.Date;
+import java.util.Properties;
+import java.util.Set;
+
+/**
+ * This class implements {@link Authorizer} from CDAP and is responsible for interacting with Ranger to enforece
```
poornachandra: `enforece` => `enforce`

cdap-ranger/cdap-ranger-binding/src/main/java/co/cask/cdap/security/authorization/ranger/binding/RangerAuthorizer.java
```diff
@@ -0,0 +1,301 @@
...
+  private static final String KEY_STREAM = "stream";
+  private static final String KEY_PROGRAM = "program";
+  private static final String KEY_DATASET_MODULE = "dataset_module";
+  private static final String KEY_DATASET_TYPE = "dataset_type";
+  private static final String KEY_SECUREKEY = "securekey";
+
+  // using # as we don't allow it in entity names
+  private static final String RESOURCE_SEPARATOR = "#";
+
+  private static volatile RangerBasePlugin rangerPlugin = null;
```
poornachandra, line 78: Why is `rangerPlugin` static, whereas `context` and `instanceName` are not?

cdap-ranger/cdap-ranger-binding/src/main/java/co/cask/cdap/security/authorization/ranger/binding/RangerAuthorizer.java
```diff
@@ -0,0 +1,301 @@
...
+  @Override
+  public synchronized void initialize(AuthorizationContext context) throws Exception {
+    this.context = context;
+    Properties properties = context.getExtensionProperties();
+    instanceName = properties.containsKey(Constants.INSTANCE_NAME) ?
+      properties.getProperty(Constants.INSTANCE_NAME) : "cdap";
+    if (rangerPlugin == null) {
+      try {
+        UserGroupInformation ugi = UserGroupInformation.getLoginUser();
+        Preconditions.checkNotNull(ugi, "UserGroupInformation is null");
```
poornachandra: Let's add some more details here. Like - `Kerberos login information is not available` or something similar

cdap-ranger/cdap-ranger-binding/src/main/java/co/cask/cdap/security/authorization/ranger/binding/RangerAuthorizer.java
poornachandra: How would we support multiple instances of CDAP on a single Ranger instance? It feels like `appId` in `RangerBasePlugin(String serviceType, String appId)` should be the instance id of CDAP.

cdap-ranger/cdap-ranger-binding/src/main/java/co/cask/cdap/security/authorization/ranger/binding/RangerAuthorizer.java
```diff
@@ -0,0 +1,301 @@
...
+    }
+
+    if (principal.getType() != Principal.PrincipalType.USER) {
+      throw new IllegalArgumentException(String.format("The principal type for current enforcement request is '%s'. " +
+                                                         "Authorization enforcement is only supported for '%s'.",
+                                                       principal.getType(), Principal.PrincipalType.USER));
+    }
+    String requestingUser = principal.getName();
+    String ip = InetAddress.getLocalHost().getHostName();
+    java.util.Set<String> userGroups = MiscUtil.getGroupsForRequestUser(requestingUser);
```
poornachandra: Full package name not required here, right?

poornachandra (commented):
Do we have to add all the xml files like `ranger-cdap-audit.xml` and `ranger-policymgr-ssl.xml` in `src/main/resources`? Since they mostly have default values, I don't see the need to add them. If they are required for testing then we can just add them to `src/test/resources`.

cdap-ranger/cdap-ranger-lookup/src/main/java/co/cask/cdap/security/authorization/ranger/lookup/RangerLookupService.java
rohitsinha54: No. This goes inside ranger. Their logging is not same as our's lets follow the convention which they do for the code which will be on ranger side.

cdap-ranger/cdap-ranger-lookup/src/main/java/co/cask/cdap/security/authorization/ranger/lookup/RangerLookupService.java
rohitsinha54: Of course

cdap-ranger/cdap-ranger-binding/src/main/resources/ranger-policymgr-ssl.xml
rohitsinha54: Not needed as all those stuff is a part of this jira.

cdap-ranger/cdap-ranger-binding/src/main/java/co/cask/cdap/security/authorization/ranger/binding/RangerAuthorizer.java
rohitsinha54: This is the plugin name. The way to support multiple instance for ranger is when a plugin is deployed and then a service is created out of it takes a name. That will differ.

rohitsinha54: This part
image

cdap-ranger/cdap-ranger-binding/src/main/java/co/cask/cdap/security/authorization/ranger/binding/RangerAuthorizer.java
rohitsinha54: I think the idea to make RangerBasePlugin static is that there should be one instance of it. It is possible that there can be multiple instance of RangerAuthorizer and even so for different instance of cdap but they will use the same base plugin object.

cdap-ranger/cdap-ranger-binding/src/main/java/co/cask/cdap/security/authorization/ranger/binding/RangerAuthorizer.java
rohitsinha54: Also there is no documentation for the parameters but from looking into usage. The first para is servicetype and another is application like CLI  or UI. As of now we don't differentiate between them and both serves request through master which talks to ranger so it should be fine for now.

rohitsinha54:
> Do we have to add all the xml files like ranger-cdap-audit.xml and ranger-policymgr-ssl.xml in src/main/resources? Since they mostly have default values, I don't see the need to add them. If they are required for testing then we can just add them to src/test/resources.

The values are unchanged but we need the files because without these the plugin initialization fails. I think there is no default value assignment if value doesn't exists so we need to provide them in the file.

cdap-ranger/cdap-ranger-binding/src/main/java/co/cask/cdap/security/authorization/ranger/binding/RangerAuthorizer.java
```diff
@@ -0,0 +1,301 @@
...
+    instanceName = properties.containsKey(Constants.INSTANCE_NAME) ?
+      properties.getProperty(Constants.INSTANCE_NAME) : "cdap";
+    if (rangerPlugin == null) {
+      try {
+        UserGroupInformation ugi = UserGroupInformation.getLoginUser();
+        Preconditions.checkNotNull(ugi, "Kerberos login information is not available. UserGroupInformation is null");
+        // set the login user as the user as whom cdap is running as this is needed for kerberos authentication
+        MiscUtil.setUGILoginUser(ugi, null);
+        LOG.debug("Initializing Ranger CDAP Plugin with UGI {}", ugi);
+      } catch (Throwable t) {
```
poornachandra: Let's just propagate this exception so that the initialize fails

cdap-ranger/cdap-ranger-binding/src/main/java/co/cask/cdap/security/authorization/ranger/binding/RangerAuthorizer.java
```diff
@@ -0,0 +1,301 @@
...
+        LOG.error("Error getting principal.", t);
+      }
+      // the string name here should not be changed as this uniquely identifies the plugin in ranger. If it's
+      // changed it will require changing all the supporting xml file which is in this package.
```
poornachandra: If we move the XML files out of `src/main/resources` then how will this change?

cdap-ranger/cdap-ranger-binding/src/main/java/co/cask/cdap/security/authorization/ranger/binding/RangerAuthorizer.java
```diff
@@ -0,0 +1,301 @@
...
+      throw new RuntimeException("CDAP Ranger Authorizer is not initialized.");
+    }
+
+    if (principal.getType() != Principal.PrincipalType.USER) {
+      throw new IllegalArgumentException(String.format("The principal type for current enforcement request is '%s'. " +
+                                                         "Authorization enforcement is only supported for '%s'.",
+                                                       principal.getType(), Principal.PrincipalType.USER));
+    }
+    String requestingUser = principal.getName();
+    String ip = InetAddress.getLocalHost().getHostName();
```
poornachandra, line 118: Looking at other Ranger plugins, this should be the client IP. Can we get the client IP here?

cdap-ranger/cdap-ranger-binding/src/main/java/co/cask/cdap/security/authorization/ranger/binding/RangerAuthorizer.java
```diff
@@ -0,0 +1,301 @@
...
+
+    if (!isAuthorized) {
+      throw new UnauthorizedException(principal, action, entity);
+    }
+  }
+
+  @Override
+  public void enforce(EntityId entityId, Principal principal, Set<Action> set) throws Exception {
+    LOG.debug("Enforce called on entity {}, principal {}, actions {}", entityId, principal, set);
+    //TODO: Investigate if its possible to make the enforce call with set of actions rather than one by one
```
poornachandra, line 164: This method in `RangerBasePlugin` can be used for multi-check - `public Collection<RangerAccessResult> isAccessAllowed(Collection<RangerAccessRequest> requests)`

cdap-ranger/cdap-ranger-binding/src/main/java/co/cask/cdap/security/authorization/ranger/binding/RangerAuthorizer.java
```diff
@@ -0,0 +1,301 @@
...
+    return null;
+  }
+
+  @Override
+  public Set<Privilege> listPrivileges(Principal principal) throws Exception {
+    LOG.warn("List privileges operation not supported by Ranger for CDAP");
+    return null;
+  }
+
+  private String getRequestingUser() throws IllegalArgumentException {
```
poornachandra: Unused private method

cdap-ranger/cdap-ranger-binding/src/main/java/co/cask/cdap/security/authorization/ranger/binding/RangerAuthorizer.java
```diff
@@ -0,0 +1,301 @@
...
+      case ARTIFACT:
+        ArtifactId artifactId = (ArtifactId) entityId;
+        setAccessResource(artifactId.getParent(), rangerAccessResource);
+        rangerAccessResource.setValue(KEY_ARTIFACT, artifactId.getArtifact() + RESOURCE_SEPARATOR +
+          artifactId.getVersion());
+        break;
+      case APPLICATION:
+        ApplicationId applicationId = (ApplicationId) entityId;
+        setAccessResource(applicationId.getParent(), rangerAccessResource);
+        rangerAccessResource.setValue(KEY_APPLICATION, applicationId.getApplication() + RESOURCE_SEPARATOR +
```
poornachandra: When an admin has to set a policy in ranger for an application, do they have to use `#` in the entity id? In Sentry we don't use application version, however we use artifact versions in Sentry. Does it make sense to have policies on application/artifact versions?

cdap-ranger/cdap-ranger-lookup/pom.xml
```diff
@@ -0,0 +1,76 @@
...
+  </dependencies>
+
+  <build>
+    <plugins>
+      <plugin>
+        <artifactId>maven-assembly-plugin</artifactId>
+        <configuration>
+          <archive>
+            <manifest>
+              <mainClass>fully.qualified.MainClass</mainClass>
```
poornachandra: This needs to be updated

cdap-ranger/cdap-ranger-lookup/pom.xml
```diff
@@ -0,0 +1,76 @@
...
+    <dependency>
+      <groupId>co.cask.cdap</groupId>
+      <artifactId>cdap-cli</artifactId>
+    </dependency>
+  </dependencies>
+
+  <build>
+    <plugins>
+      <plugin>
+        <artifactId>maven-assembly-plugin</artifactId>
```
poornachandra: `maven-bundle-plugin` is used in `cdap-ranger-binding` pom, it would be good to use the same here too.

cdap-ranger/cdap-ranger-lookup/src/main/java/co/cask/cdap/security/authorization/ranger/lookup/client/CDAPConnectionManager.java
```diff
@@ -0,0 +1,60 @@
...
+   *
+   * @param serviceName the name of the service e.g. cdap
+   * @param configs configs to use to create the cdap clients.
+   * @return {@link CDAPRangerLookupClient}
+   */
+  public static CDAPRangerLookupClient getCDAPClient(String serviceName, Map<String, String> configs) {
+    String instanceURL = configs.get(INSTANCE_URL);
+    String username = configs.get(USERNAME);
+    String password = configs.get(PASSWORD);
+    if (!(Strings.isNullOrEmpty(instanceURL) || Strings.isNullOrEmpty(username) || Strings.isNullOrEmpty(password))) {
```
poornachandra: Strictly speaking username and password are not absolutely necessary to connect to a CDAP cluster. I think we can relax this requirement, since we use the non-null status of username and password in `CDAPRangerLookupClient` to fetch access token.

cdap-ranger/cdap-ranger-lookup/src/main/java/co/cask/cdap/security/authorization/ranger/lookup/client/CDAPRangerLookupClient.java
```diff
@@ -0,0 +1,222 @@
...
+
+  private void initConnection() {
+    if (LOG.isDebugEnabled()) {
+      LOG.debug("==> CDAPRangerLookupClient initConnection()");
+    }
+    if (username != null && password != null) {
+      // security is enabled, we need to get access token before checking system services
+      try {
+        LOG.info(String.format("Fetching access token with username %s and password ****", username));
+        accessToken = fetchAccessToken(username, password);
```
poornachandra: Instead of saving the access token, if you always call `authClient.getAccessToken()` every time you need the access token, then on token expiry the auth client automatically fetches a new token.

cdap-ranger/cdap-ranger-lookup/src/main/java/co/cask/cdap/security/authorization/ranger/lookup/client/CDAPRangerLookupClient.java
```diff
@@ -0,0 +1,222 @@
...
+   */
+  private AccessToken fetchAccessToken(String username, String password) throws IOException, TimeoutException,
+    InterruptedException {
+    Properties properties = new Properties();
+    properties.setProperty("security.auth.client.username", username);
+    properties.setProperty("security.auth.client.password", password);
+    final AuthenticationClient authClient = new BasicAuthenticationClient();
+    authClient.configure(properties);
+    ConnectionConfig connectionConfig = getClientConfig().getConnectionConfig();
+    authClient.setConnectionInfo(connectionConfig.getHostname(), connectionConfig.getPort(), false);
```
poornachandra: SSL should be configurable

cdap-ranger/cdap-ranger-lookup/src/main/java/co/cask/cdap/security/authorization/ranger/lookup/client/CDAPRangerLookupClient.java
```diff
@@ -0,0 +1,222 @@
...
+    if (LOG.isDebugEnabled()) {
+      LOG.debug("==> CDAPRangerLookupClient initConnection()");
+    }
+    if (username != null && password != null) {
+      // security is enabled, we need to get access token before checking system services
+      try {
+        LOG.info(String.format("Fetching access token with username %s and password ****", username));
+        accessToken = fetchAccessToken(username, password);
+      } catch (Exception ex) {
+        throw Throwables.propagate(ex);
```
poornachandra: This will most likely throw only `IOException`, let's declare the exception and only throw that.

cdap-ranger/cdap-ranger-lookup/src/main/java/co/cask/cdap/security/authorization/ranger/lookup/client/CDAPRangerLookupClient.java
```diff
@@ -0,0 +1,222 @@
...
+      }
+    }
+    if (LOG.isDebugEnabled()) {
+      LOG.debug("<== CDAPRangerLookupClient initConnection()");
+    }
+  }
+
+  /**
+   * Tests that a connection can be established to CDAP with the given config. We do this by listing namespaces.
+   * @return Map<String, Object> Connection test response
```
poornachandra: `< >` in javadoc will be treated as an HTML tag, use `&lt;` and `&gt;` instead.

cdap-ranger/cdap-ranger-lookup/src/main/java/co/cask/cdap/security/authorization/ranger/lookup/client/CDAPRangerLookupClient.java
```diff
@@ -0,0 +1,222 @@
...
+   */
+  Map<String, Object> testConnection() {
+    if (LOG.isDebugEnabled()) {
+      LOG.debug("==> CDAPRangerLookupClient testConnection()");
+    }
+    HashMap<String, Object> responseData = new HashMap<>();
+    List<String> testResult;
+    try {
+      testResult = getNamespaces(null);
+      if (testResult.size() >= 0) {
```
poornachandra: This condition will always evaluate to true, use `!testResult.isEmpty()` instead.

cdap-ranger/cdap-ranger-lookup/src/main/java/co/cask/cdap/security/authorization/ranger/lookup/client/CDAPRangerLookupClient.java
```diff
@@ -0,0 +1,222 @@
...
+      } catch (Throwable e) {
+        // Also suppress and retry if the root cause is IOException
+        Throwable rootCause = Throwables.getRootCause(e);
+        if (!(rootCause instanceof IOException)) {
+          // Throw if root cause is any other exception e.g. UnauthenticatedException
+          throw Throwables.propagate(rootCause);
+        }
+      }
+      TimeUnit.SECONDS.sleep(1);
+    } while (System.currentTimeMillis() <= startingTime + SERVICE_TIMEOUT_SECONDS * 1000);
```
poornachandra, line 219: Connection timeout is 120s and service timeout is 30s, so this will be tried only once.

cdap-ranger/cdap-ranger-lookup/src/main/resources/ranger-servicedef-cdap.json
```diff
@@ -0,0 +1,279 @@
...
+      "level": 20,
+      "parent": "instance",
+      "mandatory": false,
+      "lookupSupported": true,
+      "recursiveSupported": false,
+      "excludesSupported": true,
+      "matcher": "org.apache.ranger.plugin.resourcematcher.RangerDefaultResourceMatcher",
+      "matcherOptions": {
+        "wildCard": true,
+        "ignoreCase": true
```
poornachandra: We should match the case of the entities too since namespace `NS` is different from namespace `ns` in CDAP.

cdap-ranger/cdap-ranger-lookup/src/main/resources/ranger-servicedef-cdap.json
```diff
@@ -0,0 +1,279 @@
...
+      },
+      "validationRegEx": "",
```
poornachandra: Would be good to have validation regex for each entity so that user gets immediate feedback

cdap-ranger/cdap-ranger-lookup/src/main/resources/ranger-servicedef-cdap.json
```diff
@@ -0,0 +1,279 @@
...
+    {
+      "itemId": 30,
+      "name": "application",
+      "type": "string",
+      "level": 30,
+      "parent": "namespace",
+      "mandatory": false,
+      "lookupSupported": true,
+      "recursiveSupported": false,
+      "excludesSupported": true,
```
poornachandra, line 59: Is there a Ranger doc that explains what these fields mean?

cdap-ranger/cdap-ranger-lookup/src/main/resources/ranger-servicedef-cdap.json
```diff
@@ -0,0 +1,279 @@
...
+      "label": "Admin"
+    }
+  ],
+  "configs": [
+    {
+      "itemId": 1,
+      "name": "cdap.username",
+      "type": "string",
+      "subType": "",
+      "mandatory": true,
```
poornachandra, line 247: Again - should username and password be mandatory?

pom.xml
```diff
@@ -121,6 +122,29 @@
         <version>${junit.version}</version>
         <scope>test</scope>
       </dependency>
+      <dependency>
+        <groupId>co.cask.cdap</groupId>
+        <artifactId>cdap-client</artifactId>
+        <version>${cdap.version}</version>
+        <!--TOOO rohit: This is probably not needed-->
```
poornachandra: Remove if not needed

poornachandra (commented):
@rohitsinha54 I did a pass, please address the comments. 

Also - I notice that the IDE is showing some warnings, please take a look at them too.

cdap-ranger/cdap-ranger-binding/src/main/java/co/cask/cdap/security/authorization/ranger/binding/RangerAuthorizer.java
rohitsinha54: Whats the rational behind it ? Are we deciding that extension initialization failure should cause cdap master start up  failure ?

cdap-ranger/cdap-ranger-binding/src/main/java/co/cask/cdap/security/authorization/ranger/binding/RangerAuthorizer.java
rohitsinha54: service type should be the part of the filename regardless of where it exists.

cdap-ranger/cdap-ranger-lookup/src/main/java/co/cask/cdap/security/authorization/ranger/lookup/client/CDAPRangerLookupClient.java
poornachandra: You can use `public ClientConfig.Builder setAccessToken(Supplier<AccessToken> accessToken)` for this.

cdap-ranger/cdap-ranger-binding/src/main/java/co/cask/cdap/security/authorization/ranger/binding/RangerAuthorizer.java
rohitsinha54: That will be done in another PR. We don't know the exact behavior of it. It needs through testing.

cdap-ranger/cdap-ranger-binding/src/main/java/co/cask/cdap/security/authorization/ranger/binding/RangerAuthorizer.java
rohitsinha54: AuthenticationContext does not have it as of now. If we need to use that we will need to do some changes in CDAP to propagate the ip here.

cdap-ranger/cdap-ranger-binding/src/main/java/co/cask/cdap/security/authorization/ranger/binding/RangerAuthorizer.java
rohitsinha54: No. They don't have it. Its stored  with this as a separator and while reading back its removed. This has been made consistent with sentry now. It now only has application name. To make it consistent with  sentry artifact still have version.

cdap-ranger/cdap-ranger-lookup/pom.xml
rohitsinha54: We need a flat uber jar for sentry. I was not able to find out how to achieve it maven-bundle-plugin. Please point to some resource to achieve it if it's possible

cdap-ranger/cdap-ranger-lookup/src/main/java/co/cask/cdap/security/authorization/ranger/lookup/client/CDAPConnectionManager.java
rohitsinha54: Authorization is meaningless without authentication in general. CDAP authorization depends on authentication to be enabled. So we are certain to talk to an authentication enabled CDAP instance.

cdap-ranger/cdap-ranger-lookup/src/main/resources/ranger-servicedef-cdap.json
rohitsinha54: I tried to find it earlier. I dont think there is.

cdap-ranger/cdap-ranger-lookup/src/main/resources/ranger-servicedef-cdap.json
rohitsinha54: Yes, we can do it later in some other PR if we have time since its a nice to have feature.

cdap-ranger/cdap-ranger-lookup/src/main/resources/ranger-servicedef-cdap.json
rohitsinha54: Done

cdap-ranger/cdap-ranger-binding/src/main/java/co/cask/cdap/security/authorization/ranger/binding/RangerAuthorizer.java
rohitsinha54: Discussed in person. We decided to fail master.

cdap-ranger/cdap-ranger-lookup/src/main/java/co/cask/cdap/security/authorization/ranger/lookup/client/CDAPRangerLookupClient.java
rohitsinha54: Actually thats what we want to test here. Even if we don't see any namespace thats fine it just mean the given user does not privilege on any of the ns. We show the message if call was successful. if it failed then we show error.

cdap-ranger/cdap-ranger-lookup/src/main/java/co/cask/cdap/security/authorization/ranger/lookup/client/CDAPRangerLookupClient.java
rohitsinha54: Actually its TimeUnit.MINUTES.toSeconds(30) so its like 30 mins represented in seconds

cdap-ranger/cdap-ranger-binding/src/main/java/co/cask/cdap/security/authorization/ranger/binding/RangerAuthorizer.java
poornachandra: It's okay for now, file a JIRA for this.

cdap-ranger/cdap-ranger-binding/src/main/java/co/cask/cdap/security/authorization/ranger/binding/RangerAuthorizer.java
poornachandra: Can you add a comment to CDAP-12148 regarding this, so that we don't miss this?

cdap-ranger/cdap-ranger-binding/src/main/java/co/cask/cdap/security/authorization/ranger/binding/RangerAuthorizer.java
poornachandra: I think we should remove versions from artifact too. Let's discuss this.

cdap-ranger/cdap-ranger-lookup/pom.xml
poornachandra: You could try exporting all classes and removing the embed dependency tag. I see you have added why you are using this plugin now. For now this is okay.

cdap-ranger/cdap-ranger-lookup/src/main/java/co/cask/cdap/security/authorization/ranger/lookup/client/CDAPRangerLookupClient.java
```diff
@@ -174,30 +177,32 @@ private ClientConfig getClientConfig() {
...
     Properties properties = new Properties();
     properties.setProperty("security.auth.client.username", username);
     properties.setProperty("security.auth.client.password", password);
-    final AuthenticationClient authClient = new BasicAuthenticationClient();
+    authClient = new BasicAuthenticationClient();
     authClient.configure(properties);
     ConnectionConfig connectionConfig = getClientConfig().getConnectionConfig();
-    authClient.setConnectionInfo(connectionConfig.getHostname(), connectionConfig.getPort(), false);
+    authClient.setConnectionInfo(connectionConfig.getHostname(), connectionConfig.getPort(),
+                                 connectionConfig.getHostname().startsWith("https://"));
```
poornachandra: Use `ConnectionConfig.isSSLEnabled()` instead

cdap-ranger/cdap-ranger-lookup/src/main/java/co/cask/cdap/security/authorization/ranger/lookup/client/CDAPRangerLookupClient.java
poornachandra: In that case why do we need the if condition?

cdap-ranger/cdap-ranger-lookup/src/main/java/co/cask/cdap/security/authorization/ranger/lookup/client/CDAPRangerLookupClient.java
poornachandra: In that case it is too much. I think we should have connection timeout of 30 seconds, and total timeout of 90 or 120 seconds. What do you say?

cdap-ranger/cdap-ranger-lookup/src/main/resources/ranger-servicedef-cdap.json
poornachandra: Can you add a comment to JIRA CDAP-12148, so that we don't miss this?

cdap-ranger/cdap-ranger-lookup/src/main/resources/ranger-servicedef-cdap.json
poornachandra: Let's try to figure this out. Otherwise how do you make the decision whether this has to be true or false?

poornachandra (commented):
A few more comments to address.

poornachandra:
Also - it would be good if you add some tests

cdap-ranger/cdap-ranger-binding/src/main/java/co/cask/cdap/security/authorization/ranger/binding/RangerAuthorizer.java
rohitsinha54: https://issues.cask.co/browse/CDAP-12264

cdap-ranger/cdap-ranger-binding/src/main/java/co/cask/cdap/security/authorization/ranger/binding/RangerAuthorizer.java
rohitsinha54: Done.

cdap-ranger/cdap-ranger-lookup/src/main/java/co/cask/cdap/security/authorization/ranger/lookup/client/CDAPRangerLookupClient.java
rohitsinha54: Right.

cdap-ranger/cdap-ranger-lookup/src/main/java/co/cask/cdap/security/authorization/ranger/lookup/client/CDAPRangerLookupClient.java
rohitsinha54: Ya I think 5 retries should be fine.

poornachandra (approved):

PR #134, opened 2017-08-09
CDAP-12148 Add lookup for all CDAP entities

rohitsinha54:
- Adds lookup client for cdap to support resource lookup from Ranger.

### DIFF
--- a/cdap-ranger/cdap-ranger-binding/pom.xml
+++ b/cdap-ranger/cdap-ranger-binding/pom.xml
@@ -58,6 +58,11 @@
       <groupId>org.apache.ranger</groupId>
       <artifactId>ranger-plugins-audit</artifactId>
     </dependency>
+    <dependency>
+      <groupId>co.cask.cdap</groupId>
+      <artifactId>cdap-ranger-common</artifactId>
+      <version>${project.version}</version>
+    </dependency>
   </dependencies>
 
   <build>
--- a/cdap-ranger/cdap-ranger-lookup/src/main/java/co/cask/cdap/security/authorization/ranger/lookup/client/CDAPRangerLookupClient.java
+++ b/cdap-ranger/cdap-ranger-lookup/src/main/java/co/cask/cdap/security/authorization/ranger/lookup/client/CDAPRangerLookupClient.java
@@ -15,18 +15,41 @@
  */
 package co.cask.cdap.security.authorization.ranger.lookup.client;
 
+import co.cask.cdap.api.artifact.ArtifactSummary;
 import co.cask.cdap.cli.util.InstanceURIParser;
+import co.cask.cdap.client.ApplicationClient;
+import co.cask.cdap.client.ArtifactClient;
+import co.cask.cdap.client.DatasetClient;
+import co.cask.cdap.client.DatasetModuleClient;
+import co.cask.cdap.client.DatasetTypeClient;
 import co.cask.cdap.client.NamespaceClient;
+import co.cask.cdap.client.SecureStoreClient;
+import co.cask.cdap.client.StreamClient;
 import co.cask.cdap.client.config.ClientConfig;
 import co.cask.cdap.client.config.ConnectionConfig;
+import co.cask.cdap.proto.ApplicationRecord;
+import co.cask.cdap.proto.DatasetModuleMeta;
+import co.cask.cdap.proto.DatasetSpecificationSummary;
+import co.cask.cdap.proto.DatasetTypeMeta;
 import co.cask.cdap.proto.NamespaceMeta;
+import co.cask.cdap.proto.ProgramRecord;
+import co.cask.cdap.proto.StreamDetail;
+import co.cask.cdap.proto.id.ApplicationId;
+import co.cask.cdap.proto.id.NamespaceId;
 import co.cask.cdap.security.authentication.client.AccessToken;
 import co.cask.cdap.security.authentication.client.AuthenticationClient;
 import co.cask.cdap.security.authentication.client.basic.BasicAuthenticationClient;
+import co.cask.cdap.security.authorization.ranger.commons.RangerCommon;
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Joiner;
+import com.google.common.base.Preconditions;
+import com.google.common.base.Supplier;
 import com.google.common.base.Throwables;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 import org.apache.ranger.plugin.client.BaseClient;
+import org.apache.ranger.plugin.service.ResourceLookupContext;
+import org.apache.ranger.plugin.util.TimedEventUtil;
 
 import java.io.IOException;
 import java.net.URI;
@@ -49,29 +72,53 @@ import javax.annotation.Nullable;
 public class CDAPRangerLookupClient {
 
   private static final Log LOG = LogFactory.getLog(CDAPRangerLookupClient.class);
-  private static final long SERVICE_TIMEOUT_SECONDS = TimeUnit.MINUTES.toSeconds(10);
-
+  private static final long SERVICE_TIMEOUT_SECONDS = TimeUnit.MINUTES.toSeconds(3);
   private static final String ERR_MSG = " You can still save the repository and start creating "
     + "policies, but you would not be able to use autocomplete for "
     + "resource names. Check xa_portal.log for more info.";
-  private AccessToken accessToken;
-  private final String serviceName;
   private final String instanceURL;
   private final String username;
   private final String password;
   private AuthenticationClient authClient;
-  private final NamespaceClient nsClient;
+  private NamespaceClient nsClient;
+  private StreamClient streamClient;
+  private ApplicationClient applicationClient;
+  private DatasetClient datasetClient;
+  private ArtifactClient artifactClient;
+  private DatasetModuleClient datasetModuleClient;
+  private DatasetTypeClient datasetTypeClient;
+  private SecureStoreClient secureStoreClient;
 
-  CDAPRangerLookupClient(String serviceName, String instanceURL,
-                         String username, String password) throws IOException, TimeoutException, InterruptedException {
-    this.serviceName = serviceName;
+  CDAPRangerLookupClient(String instanceURL, String username, String password) throws IOException {
+    this(instanceURL, username, password, true);
+  }
+
+  // for testing only. does not use authentication client
+  @VisibleForTesting
+  CDAPRangerLookupClient(String instanceURL, String username, String password,
+                         NamespaceClient nsClient, StreamClient streamClient, ApplicationClient applicationClient,
+                         DatasetClient datasetClient, ArtifactClient artifactClient,
+                         DatasetModuleClient datasetModuleClient, DatasetTypeClient datasetTypeClient,
+                         SecureStoreClient secureStoreClient) throws IOException {
+    this(instanceURL, username, password, false);
+    this.nsClient = nsClient;
+    this.streamClient = streamClient;
+    this.applicationClient = applicationClient;
+    this.datasetClient = datasetClient;
+    this.artifactClient = artifactClient;
+    this.datasetModuleClient = datasetModuleClient;
+    this.datasetTypeClient = datasetTypeClient;
+    this.secureStoreClient = secureStoreClient;
+  }
+
+  private CDAPRangerLookupClient(String instanceURL, String username, String password, boolean initClients) throws
+    IOException {
     this.instanceURL = instanceURL;
     this.username = username;
     this.password = password;
-    initConnection();
-    ClientConfig clientConfig = getClientConfig();
-    this.nsClient = new NamespaceClient(clientConfig);
-    //TODO create more cdap clients here
+    if (initClients) {
+      initConnection();
+    }
   }
 
   private void initConnection() throws IOException {
@@ -83,6 +130,15 @@ public class CDAPRangerLookupClient {
       try {
         LOG.info(String.format("Fetching access token with username %s and password ****", username));
         initAuthClient(username, password);
+        ClientConfig clientConfig = getClientConfig();
+        this.nsClient = new NamespaceClient(clientConfig);
+        this.streamClient = new StreamClient(clientConfig);
+        this.datasetClient = new DatasetClient(clientConfig);
+        this.applicationClient = new ApplicationClient(clientConfig);
+        this.artifactClient = new ArtifactClient(clientConfig);
+        this.datasetModuleClient = new DatasetModuleClient(clientConfig);
+        this.datasetTypeClient = new DatasetTypeClient(clientConfig);
+        this.secureStoreClient = new SecureStoreClient(clientConfig);
       } catch (Exception ex) {
         Throwables.propagateIfInstanceOf(ex, IOException.class);
         throw Throwables.propagate(ex);
@@ -95,6 +151,7 @@ public class CDAPRangerLookupClient {
 
   /**
    * Tests that a connection can be established to CDAP with the given config. We do this by listing namespaces.
+   *
    * @return Map &lt; String, Object &gt; Connection test response
    */
   Map<String, Object> testConnection() {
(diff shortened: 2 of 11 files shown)


cdap-ranger/cdap-ranger-binding/pom.xml
```diff
@@ -58,6 +58,11 @@
       <groupId>org.apache.ranger</groupId>
       <artifactId>ranger-plugins-audit</artifactId>
     </dependency>
+    <dependency>
+      <groupId>co.cask.cdap</groupId>
+      <artifactId>cdap-ranger-commons</artifactId>
```
poornachandra: Let's call the module `cdap-ranger-common`

cdap-ranger/cdap-ranger-commons/pom.xml
```diff
@@ -0,0 +1,39 @@
...
+  <parent>
+    <artifactId>cdap-ranger</artifactId>
+    <groupId>co.cask.cdap</groupId>
+    <version>0.1.0-SNAPSHOT</version>
+  </parent>
+  <modelVersion>4.0.0</modelVersion>
+
+  <artifactId>cdap-ranger-commons</artifactId>
+
+  <name>CDAP Ranger Commons</name>
```
poornachandra: Let's replace commons with common everywhere.

cdap-ranger/cdap-ranger-lookup/src/main/java/co/cask/cdap/security/authorization/ranger/lookup/client/CDAPRangerLookupClient.java
```diff
@@ -119,6 +176,84 @@ private void initConnection() throws IOException {
...
+            List<String> retList = new ArrayList<>();
+            try {
+              List<String> list = null;
+              if (resource.trim().equalsIgnoreCase(RangerCommons.KEY_NAMESPACE)) {
+                list = getNamespaces(resourceMap.get(RangerCommons.KEY_NAMESPACE));
+              } else {
+                // All of these belong to a namespace
+                Preconditions.checkNotNull(resourceMap.get(RangerCommons.KEY_NAMESPACE));
+                NamespaceId namespace = new NamespaceId(resourceMap.get(RangerCommons.KEY_NAMESPACE).get(0));
+                if (resource.trim().equalsIgnoreCase(RangerCommons.KEY_STREAM)) {
```
poornachandra: Replacing if-else ladder with switch will improve readability

cdap-ranger/cdap-ranger-lookup/src/main/java/co/cask/cdap/security/authorization/ranger/lookup/client/CDAPRangerLookupClient.java
```diff
@@ -119,6 +176,84 @@ private void initConnection() throws IOException {
...
+                } else if (resource.trim().equalsIgnoreCase(RangerCommons.KEY_DATASET_TYPE)) {
+                  list = getDatasetTypes(resourceMap.get(RangerCommons.KEY_DATASET_TYPE), namespace);
+                } else if (resource.trim().equalsIgnoreCase(RangerCommons.KEY_SECUREKEY)) {
+                  list = getSecureKeys(resourceMap.get(RangerCommons.KEY_SECUREKEY), namespace);
+                }
+              }
+              Preconditions.checkNotNull(list, "Failed to list resources of type %s", resource.trim());
+              if (!userInput.isEmpty()) {
+                for (String value : list) {
+                  if (value.startsWith(userInput)) {
```
poornachandra, line 238: As an enhancement we could use the search APIs instead of the clients to get the list of entities (except secure keys). That way we don't have to list everything and then filter, instead can directly send the search term to the backend.

cdap-ranger/cdap-ranger-lookup/src/main/java/co/cask/cdap/security/authorization/ranger/lookup/client/CDAPRangerLookupClient.java
```diff
@@ -119,6 +176,84 @@ private void initConnection() throws IOException {
...
+                retList.addAll(list);
+              }
+            } catch (Exception ex) {
+              LOG.error("Error getting resource.", ex);
+            }
+            return retList;
+          }
+        };
+
+        synchronized (this) {
```
poornachandra, line 252: Why is this synchronized block needed?

cdap-ranger/cdap-ranger-lookup/src/main/java/co/cask/cdap/security/authorization/ranger/lookup/client/CDAPRangerLookupClient.java
```diff
@@ -119,6 +176,84 @@ private void initConnection() throws IOException {
...
+          resultList = TimedEventUtil.timedTask(callableObj, SERVICE_TIMEOUT_SECONDS, TimeUnit.SECONDS);
```
poornachandra, line 253: If we really need to timeout the call then `TimedEventUtil.timedTask()` is not useful, since the method just calls the callable directly without applying any timeouts.

cdap-ranger/cdap-ranger-lookup/src/main/java/co/cask/cdap/security/authorization/ranger/lookup/client/CDAPRangerLookupClient.java
```diff
@@ -49,29 +72,53 @@
...
+  private ArtifactClient artifactClient;
+  private DatasetModuleClient datasetModuleClient;
+  private DatasetTypeClient datasetTypeClient;
+  private SecureStoreClient secureStoreClient;
+
+  CDAPRangerLookupClient(String instanceURL, String username, String password) throws IOException {
+    this(instanceURL, username, password, true);
+  }
+
+  // for testing only. does not uses authentication client
```
poornachandra: uses => use

cdap-ranger/cdap-ranger-lookup/src/main/java/co/cask/cdap/security/authorization/ranger/lookup/client/CDAPRangerLookupClient.java
```diff
@@ -142,25 +277,208 @@ private void initConnection() throws IOException {
...
+
+    if (LOG.isDebugEnabled()) {
+      LOG.debug("<== CDAPRangerLookupClient.getStreams(): " + streams);
+    }
+    return streams;
+  }
+
+  private List<String> getDatasets(@Nullable List<String> datasetList, NamespaceId namespace) throws Exception {
+    if (LOG.isDebugEnabled()) {
+      LOG.debug("==> CDAPRangerLookupClient.getDatasets() ExcludeDatasetList :" + datasetList);
```
poornachandra: Is this `ExcludeDatasetList` or include?

cdap-ranger/cdap-ranger-lookup/src/main/java/co/cask/cdap/security/authorization/ranger/lookup/client/CDAPRangerLookupClient.java
```diff
@@ -142,25 +277,208 @@ private void initConnection() throws IOException {
...
+      LOG.warn("Failed to get Stream. StreamClient is not initialized.");
+    }
+
+    if (LOG.isDebugEnabled()) {
+      LOG.debug("<== CDAPRangerLookupClient.getStreams(): " + streams);
+    }
+    return streams;
+  }
+
+  private List<String> getDatasets(@Nullable List<String> datasetList, NamespaceId namespace) throws Exception {
```
poornachandra: Readability improves when nullable parameters are at the end, like -
```
private List<String> getDatasets(NamespaceId namespace, @Nullable List<String> datasetList)
```

cdap-ranger/cdap-ranger-lookup/src/main/java/co/cask/cdap/security/authorization/ranger/lookup/client/CDAPRangerLookupClient.java
```diff
@@ -119,6 +176,84 @@ private void initConnection() throws IOException {
...
+      try {
+        Callable<List<String>> callableObj;
+        callableObj = new Callable<List<String>>() {
+          @Override
+          public List<String> call() {
+            List<String> retList = new ArrayList<>();
+            try {
+              List<String> list = null;
+              if (resource.trim().equalsIgnoreCase(RangerCommons.KEY_NAMESPACE)) {
+                list = getNamespaces(resourceMap.get(RangerCommons.KEY_NAMESPACE));
```
poornachandra: What is the contract for `resourceMap`? In some cases we use the value as a positive filter, in some cases we use it as a negative filter.

cdap-ranger/cdap-ranger-lookup/src/main/java/co/cask/cdap/security/authorization/ranger/lookup/client/CDAPRangerLookupClient.java
```diff
@@ -142,25 +277,208 @@ private void initConnection() throws IOException {
...
+      LOG.debug("==> CDAPRangerLookupClient.getArtifacts() ExcludeArtifactList :" + artifactList);
+    }
+
+    List<String> artifacts = new ArrayList<>();
+    if (artifactClient != null) {
+      for (ArtifactSummary artifactSummary : artifactClient.list(namespace)) {
+        String name = artifactSummary.getName();
+        String version = artifactSummary.getVersion();
+        if (artifactList == null || !artifacts.contains(name)) {
+          artifacts.add(name + "(Version: " + version + ")");
```
poornachandra: Version can be removed, right? https://issues.cask.co/browse/CDAP-12283

poornachandra (commented):
@rohitsinha54 Just a few comments

cdap-ranger/cdap-ranger-lookup/src/main/java/co/cask/cdap/security/authorization/ranger/lookup/client/CDAPRangerLookupClient.java
rohitsinha54: What do you mean by positive filter and negative filter ?

cdap-ranger/cdap-ranger-lookup/src/main/java/co/cask/cdap/security/authorization/ranger/lookup/client/CDAPRangerLookupClient.java
rohitsinha54: I can do it in the follow up pr for the improvement.

cdap-ranger/cdap-ranger-lookup/src/main/java/co/cask/cdap/security/authorization/ranger/lookup/client/CDAPRangerLookupClient.java
rohitsinha54: Noted down but  the api takes that and doesn't use it.

cdap-ranger/cdap-ranger-lookup/src/main/java/co/cask/cdap/security/authorization/ranger/lookup/client/CDAPRangerLookupClient.java
rohitsinha54: I think the idea is resultSet will be created in threadlock as getResources can be called from multiple threads.

cdap-ranger/cdap-ranger-lookup/src/main/java/co/cask/cdap/security/authorization/ranger/lookup/client/CDAPRangerLookupClient.java
rohitsinha54: Exclude

cdap-ranger/cdap-ranger-lookup/src/main/java/co/cask/cdap/security/authorization/ranger/lookup/client/CDAPRangerLookupClient.java
rohitsinha54: We decided we will have version when the code was written. The requirement changed now. I will update it now.

cdap-ranger/cdap-ranger-lookup/src/main/java/co/cask/cdap/security/authorization/ranger/lookup/client/CDAPRangerLookupClient.java
poornachandra: File a JIRA, we can do it later

cdap-ranger/cdap-ranger-lookup/src/main/java/co/cask/cdap/security/authorization/ranger/lookup/client/CDAPRangerLookupClient.java
poornachandra: That's what I'm trying to understand - why we need result set to be created in a synchronized block. All our clients are thread-safe, right? In that case I don't think the synchronized block is needed. What do you say?

cdap-ranger/cdap-ranger-lookup/src/main/java/co/cask/cdap/security/authorization/ranger/lookup/client/CDAPRangerLookupClient.java
poornachandra: In line 198 and line 216 we use `resourceMap.get()` value as an inclusive filter (i.e., return only those entities that have the same namespace id or application id). However, in lines 205, 208, 211, etc. we use the `resourceMap.get()` value as an exclusion.

cdap-ranger/cdap-ranger-lookup/src/main/java/co/cask/cdap/security/authorization/ranger/lookup/client/CDAPRangerLookupClient.java
poornachandra: Hence my comment https://github.com/caskdata/cdap-security-extn/pull/134#discussion_r133127683

cdap-ranger/cdap-ranger-lookup/src/main/java/co/cask/cdap/security/authorization/ranger/lookup/client/CDAPRangerLookupClient.java
poornachandra: Also - look at my comment https://github.com/caskdata/cdap-security-extn/pull/134#discussion_r133126799

cdap-ranger/cdap-ranger-lookup/src/main/java/co/cask/cdap/security/authorization/ranger/lookup/client/CDAPRangerLookupClient.java
poornachandra: 👍

cdap-ranger/cdap-ranger-lookup/src/main/java/co/cask/cdap/security/authorization/ranger/lookup/client/CDAPRangerLookupClient.java
```diff
@@ -402,9 +410,8 @@ private void initConnection() throws IOException {
     if (artifactClient != null) {
       for (ArtifactSummary artifactSummary : artifactClient.list(namespace)) {
         String name = artifactSummary.getName();
-        String version = artifactSummary.getVersion();
         if (artifactList == null || !artifacts.contains(name)) {
-          artifacts.add(name + "(Version: " + version + ")");
+          artifacts.add(name);
```
poornachandra: `RangerAuthorizer.setAccessResource()` also has to change for artifact, right?

poornachandra (commented):
@rohitsinha54 A few more comments. 

Note some comments may be in the collapsed blocks.

cdap-ranger/cdap-ranger-lookup/src/main/java/co/cask/cdap/security/authorization/ranger/lookup/client/CDAPRangerLookupClient.java
rohitsinha54: https://issues.cask.co/browse/CDAP-12338

cdap-ranger/cdap-ranger-lookup/src/main/java/co/cask/cdap/security/authorization/ranger/lookup/client/CDAPRangerLookupClient.java
rohitsinha54: There can be multiple user using ranger UI at the same time making request to lookup resource with different prefix for different entity type. I think that the synchronized is to handle that. Although I think we can remove it but we will not be certain of the behavior unless we investigate.

cdap-ranger/cdap-ranger-lookup/src/main/java/co/cask/cdap/security/authorization/ranger/lookup/client/CDAPRangerLookupClient.java
rohitsinha54: 👍

poornachandra (approved):

PR #135, opened 2017-08-16
CDAP-12148 Support isVisible check

rohitsinha54:
Support visibility check in ranger authorizer

### DIFF
--- a/cdap-ranger/cdap-ranger-binding/src/main/java/co/cask/cdap/security/authorization/ranger/binding/RangerAuthorizer.java
+++ b/cdap-ranger/cdap-ranger-binding/src/main/java/co/cask/cdap/security/authorization/ranger/binding/RangerAuthorizer.java
@@ -41,15 +41,18 @@ import com.google.common.base.Preconditions;
 import org.apache.hadoop.security.UserGroupInformation;
 import org.apache.ranger.audit.provider.MiscUtil;
 import org.apache.ranger.plugin.audit.RangerDefaultAuditHandler;
+import org.apache.ranger.plugin.policyengine.RangerAccessRequest;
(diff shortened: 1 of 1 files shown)


cdap-ranger/cdap-ranger-binding/src/main/java/co/cask/cdap/security/authorization/ranger/binding/RangerAuthorizer.java
```diff
@@ -215,6 +179,61 @@ public void removeRoleFromPrincipal(Role role, Principal principal) throws Excep
...
+    }
+
+    if (principal.getType() != Principal.PrincipalType.USER) {
+      throw new IllegalArgumentException(String.format("The principal type for current enforcement request is '%s'. " +
+                                                         "Authorization enforcement is only supported for '%s'.",
+                                                       principal.getType(), Principal.PrincipalType.USER));
+    }
+    String requestingUser = principal.getName();
+    String ip = InetAddress.getLocalHost().getHostName();
+    Set<String> userGroups = MiscUtil.getGroupsForRequestUser(requestingUser);
```
poornachandra, line 198: When doing multiple enforce requests as part of isVisible, we can compute all these including date only once per request. Since you already mentioned that you want to use batch requests later, we can make this change at that time.

poornachandra (approved):

cdap-ranger/cdap-ranger-binding/src/main/java/co/cask/cdap/security/authorization/ranger/binding/RangerAuthorizer.java
rohitsinha54: 👍

PR #149, opened 2017-08-26
Bump cdap dependency to 4.3.0

anwar6953:
Also from -SNAPSHOT from the project version.

### DIFF
--- a/cdap-security-extensions-dist/pom.xml
+++ b/cdap-security-extensions-dist/pom.xml
@@ -21,7 +21,7 @@
   <parent>
     <artifactId>cdap-security-extensions</artifactId>
     <groupId>co.cask.cdap</groupId>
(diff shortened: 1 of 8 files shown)


cdap-security-extensions-dist/pom.xml
```diff
@@ -21,7 +21,7 @@
   <parent>
     <artifactId>cdap-security-extensions</artifactId>
     <groupId>co.cask.cdap</groupId>
-    <version>0.6.0-SNAPSHOT</version>
+    <version>0.6</version>
```
sagarkapare: nit: need to be `0.6.0` everywhere in this file

sagarkapare:
one comment about changing version to `0.6.0` fix and merge

sagarkapare:
LGTM 👍

PR #169, opened 2018-09-20
Update versions in multi-module

sreevatsanraman:
Update correct versions on all modules https://issues.cask.co/browse/CDAP-14349


albertshau (approved):
lgtm

PR #179, opened 2019-04-04
Package rename to io.cdap and upgrade to 6.0.0

anew:
title says it all. Two commits best reviewed independenty:
- renaming
- adapting to 6.0.0 CDAP changes


anwar6953:
I'm curious - Does compilation fail if you don't do the second commit?

anew:
checkstyle fails with some imports out of order. Other than that, compilation went through but tests failed, because flows and streams do not exist any more, so, for example, ProgramType.valueOf("FLOW") throws exception.

anew:
Actually compilation also failed because some APIs had changed, most notably, the AuthorizationContext implements Admin and SecureStore, which have methods added and/or renamed.

anwar6953:
Got it. We should've done second commit a while back when we did the stream/flow removal, but missed it. Thanks for including it now.

anwar6953 (approved):

anew:
Right :) Thanks, so I will not squash the commits but keep'em separate.

PR #184, opened 2019-09-10
Fix the broken build

chtyim:
- Remove unnecessary dependencies on cdap-cli
-- CDAP cli is a fat-jar, causing shade plugin issue
- Fix the test resources filtering to avoid creating a strange directory outside of target directory after running test


anwar6953 (approved):

chtyim:
Build passed https://builds.cask.co/browse/CDAP-CSI96-1

PR #221, opened 2021-03-25
Bump snapshot version

rmstar:


albertshau (commented):
I don't think this project ever really changes. We can probably just keep it at 0.13.0 and bump to 0.13.1 if needed later. Or is there some build reason it needs to be bumped?

albertshau (approved):

supriyapatil-google (approved):
Thanks!

rmstar:
Yeah there's no reason to bump this; I'll close this PR.

Edit: Thinking about this a bit more, I think I'll keep this change. While it's not really necessary to bump the version when there are no changes, it does help the release mgr keep track of branch mapping. e.g CDAP 6.3 depends on release/0.12 in cdap-security-extn => CDAP 6.4 depends on release/0.13 in cdap-security-extn (increment minor version). Bumping version on every CDAP release also makes it easier to write release automation scripts.

PR #243, opened 2022-03-17
[CDAP-18928] Add support for InstancePermission.TETHER and InstancePermission.HEALTH_CHECK

dli357:
These permissions were added in https://github.com/cdapio/cdap/pull/13769 (TETHER) and https://github.com/cdapio/cdap/pull/13967 (HEALTH_CHECK).

Permissions Descriptions:
 - TETHER permission is required for the given user to initiate a tether request to a remote instance. TETHER permission is also required for a remote instance initiating a request to the local instance.
 -  HEALTH_CHECK permission is required for acquiring JVM information from a particular service.

### DIFF
--- a/cdap-ldap-role/pom.xml
+++ b/cdap-ldap-role/pom.xml
@@ -21,7 +21,7 @@
 
   <properties>
     <security.authorizer.class>io.cdap.cdap.security.authorization.ldap.role.LDAPRoleAccessController</security.authorizer.class>
-    <cdap.version>6.5.0</cdap.version>
+    <cdap.version>6.7.0-SNAPSHOT</cdap.version>
     <jackson.version>2.8.8</jackson.version>
     <maven.build.timestamp.format>HH:mm:ss dd-MM-yyyy</maven.build.timestamp.format>
   </properties>
--- a/cdap-ldap-role/src/main/java/io/cdap/cdap/security/authorization/ldap/role/permission/RolePermission.java
+++ b/cdap-ldap-role/src/main/java/io/cdap/cdap/security/authorization/ldap/role/permission/RolePermission.java
@@ -108,4 +108,10 @@ public enum RolePermission {
   MANAGE_SYSTEM_PREFERENCES,
   @JsonProperty("View System Services")
   VIEW_SYSTEM_SERVICES,
+
+  // System Administration
+  @JsonProperty("Initiate And Accept Tether")
+  INITIATE_AND_ACCEPT_TETHER,
+  @JsonProperty("Perform Health Check")
+  PERFORM_HEALTH_CHECK,
 }
--- a/cdap-ldap-role/src/main/java/io/cdap/cdap/security/authorization/ldap/role/permission/RolePermissionConverter.java
+++ b/cdap-ldap-role/src/main/java/io/cdap/cdap/security/authorization/ldap/role/permission/RolePermissionConverter.java
@@ -17,7 +17,9 @@
 package io.cdap.cdap.security.authorization.ldap.role.permission;
 
 import io.cdap.cdap.proto.element.EntityType;
+import io.cdap.cdap.proto.id.NamespacedEntityId;
 import io.cdap.cdap.proto.security.ApplicationPermission;
+import io.cdap.cdap.proto.security.InstancePermission;
 import io.cdap.cdap.proto.security.Permission;
 import io.cdap.cdap.proto.security.StandardPermission;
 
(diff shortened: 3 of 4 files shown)


rahul1995 (approved):

achernatsky-cybervision:
@dli357 You need to define also _EntityType_ in _getEntityType_ function for these permissions.

dli357:
@achernatsky-cybervision Thanks for the review! Added it to the `getEntityType` function.

Also added a unit test to ensure all permissions are validly converted from a single collection.

Open question: Does the extension attempt to map all instance-wide entities to the system namespace? Unsure if returning `EntityType.INSTANCE` may cause issues or not.

achernatsky-cybervision:
If EntityId does not have namespace attribute, like InstanceId, as it does not extends NamespacedEntityId class, than it will be mapped to system namespace. [It handles here](https://github.com/cdapio/cdap-security-extn/blob/35debe2651e24aa82c66b53fd9bf977ddedd9f1d/cdap-ldap-role/src/main/java/io/cdap/cdap/security/authorization/ldap/role/group/PrincipalPermissions.java#L104-L111)
Otherwise it is necessary to specify system name in files with roles mapping or the custom mapping can be added in _convertToEntityTypeWithPermission_ function.

achernatsky-cybervision:
But it handles system namespace only for permissions request. For converting permissions I think it will be necessary to add a condition [here](https://github.com/cdapio/cdap-security-extn/blob/35debe2651e24aa82c66b53fd9bf977ddedd9f1d/cdap-ldap-role/src/main/java/io/cdap/cdap/security/authorization/ldap/role/permission/RolePermissionConverter.java#L171).

dli357:
Gotcha, so in this case it should be okay not to add it in [here](https://github.com/cdapio/cdap-security-extn/blob/35debe2651e24aa82c66b53fd9bf977ddedd9f1d/cdap-ldap-role/src/main/java/io/cdap/cdap/security/authorization/ldap/role/group/PrincipalPermissions.java#L104-L111) because it will work as intended if you manually specify these permissions in the system namespace?

These permissions are meant for the equivalent of an instance admin to perform and shouldn't be applied on a per-namespace basis.

achernatsky-cybervision:
Yes, everything should work correctly, if you specify system namespace in roles mapping file.
The result permissions are stored in Map, where key is name of namespace, that's why it is necessary to add each permission to namespace. 
It would be better to change [this line](https://github.com/cdapio/cdap-security-extn/blob/35debe2651e24aa82c66b53fd9bf977ddedd9f1d/cdap-ldap-role/src/main/java/io/cdap/cdap/security/authorization/ldap/role/permission/RolePermissionConverter.java#L171) to:
`boolean isSystemNamespace = !NamespacedEntityId.class.isAssignableFrom(entityType.getIdClass()) || permission == RolePermission.USE_STUDIO;`
It will avoid necessity  to grant access to many objects inside roles mapping file by specifying system namespace.

dli357:
@achernatsky-cybervision If I understand correctly, adding the line you mentioned would mean that you could assign the `INITIATE_AND_ACCEPT_TETHER` permission to any namespace and it would properly grant a user permission to initiate a tether across the entire instance. Are we sure this is intended behavior?

Permissions like `PERFORM_HEALTH_CHECK` and `INITIATE_AND_ACCEPT_TETHER` are meant to be administrator permissions, and administrator permissions seem like something we would want explicitly assigned to the instance rather than to a particular namespace.

It's no problem if that's the intended behavior, just want to make sure that sounds right, thanks!

achernatsky-cybervision:
Yes it will grant permissions _PERFORM_HEALTH_CHECK_ and _INITIATE_AND_ACCEPT_TETHER_ automatically to system namespace, no matter what namespace is specified. The same behavior is now used [here](https://github.com/cdapio/cdap-security-extn/blob/35debe2651e24aa82c66b53fd9bf977ddedd9f1d/cdap-ldap-role/src/main/java/io/cdap/cdap/security/authorization/ldap/role/permission/RolePermissionConverter.java#L121-L124) and [here](https://github.com/cdapio/cdap-security-extn/blob/35debe2651e24aa82c66b53fd9bf977ddedd9f1d/cdap-ldap-role/src/main/java/io/cdap/cdap/security/authorization/ldap/role/permission/RolePermissionConverter.java#L128-L129) and [here](https://github.com/cdapio/cdap-security-extn/blob/35debe2651e24aa82c66b53fd9bf977ddedd9f1d/cdap-ldap-role/src/main/java/io/cdap/cdap/security/authorization/ldap/role/permission/RolePermissionConverter.java#L135-L136) and [here](https://github.com/cdapio/cdap-security-extn/blob/35debe2651e24aa82c66b53fd9bf977ddedd9f1d/cdap-ldap-role/src/main/java/io/cdap/cdap/security/authorization/ldap/role/permission/RolePermissionConverter.java#L151-L152).

The idea is to minimize necessity for user to specify system namespace in mapping file to avoid situations when user adds the difference set of permissions and then grant them to system namespace even if some of this permissions do not require this namespace. 

I understand that PERFORM_HEALTH_CHECK and INITIATE_AND_ACCEPT_TETHER are administrator permissions and that they are not related to namespaces. But the design of storing permissions is made as Map with namespace key, as I've mentioned previously. So to make EntityIds without namespace information fit this structure I've added them to system namespace, as for example to access SYSTEM_SERVICE it is necessary to access system namespace.

dli357:
SGTM, I've gone ahead and modified the PR. PTAL, thanks!

PR #258, opened 2022-10-20
Upgrade cdap version due to security vulnerability

itsankit-google:
- Update cdap version to `6.9.0-SNAPSHOT` due to security vulnerability fixes.
- Upgraded `slf4j.version` required by `logbackv1.2.11`.
- Remove `cdap.version` from `cdap-ldap-role` so that it inherits from parent.
- Updated `Authorizable` according to changed definition in `cdap` repo.
- Excluded `log4j:log4j` & `slf4j-log4j12` from `org.apache.ranger:ranger-plugins-common`.
- Added `log4j-over-slf4j` which works as a facade.

**Testing**
- Verified using `mvn dependency:tree`.
- Build successful using `mvn clean install`.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -66,8 +66,8 @@
 
   <properties>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-    <cdap.version>6.2.0</cdap.version>
-    <slf4j.version>1.7.5</slf4j.version>
+    <cdap.version>6.9.0-SNAPSHOT</cdap.version>
+    <slf4j.version>1.7.15</slf4j.version>
     <junit.version>4.11</junit.version>
     <surefire.redirectTestOutputToFile>true</surefire.redirectTestOutputToFile>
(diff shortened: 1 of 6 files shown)


pom.xml
```diff
@@ -66,8 +66,8 @@
 
   <properties>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-    <cdap.version>6.2.0</cdap.version>
-    <slf4j.version>1.7.5</slf4j.version>
+    <cdap.version>6.9.0-SNAPSHOT</cdap.version>
```
dli357, line 69: We may need to maintain support for older versions. Any chance we can set this to 6.6?

itsankit-google: Logback & Log4j changes are there in `cdap 6.8`, so least we can use is `6.8`. Do we not have separate release branches in this repo?

dli357 (approved):

PR #295, opened 2024-06-28
CDAP-21027 : Upgrading hadoop version to 3.3.6 

sahusanket:
- Upgrading to Hadoop 3 
- This makes current cdap-sentry incompatible 
- apache Sentry is already retired since 2018 
- Removing Sentry module from compilation , but removed all vulnerability libs for safety . 
- https://cdap.atlassian.net/browse/CDAP-21050 , to be evaluated.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -25,7 +25,7 @@
 
   <modules>
     <module>cdap-ranger</module>
-    <module>cdap-sentry</module>
+<!-- TODO   <module>cdap-sentry</module> TO EVALUATE https://cdap.atlassian.net/browse/CDAP-21050 -->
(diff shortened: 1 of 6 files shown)


tivv (approved):

pom.xml
```diff
@@ -25,7 +25,7 @@
 
   <modules>
     <module>cdap-ranger</module>
-    <module>cdap-sentry</module>
+<!--    <module>cdap-sentry</module> TO EVALUATE https://cdap.atlassian.net/browse/CDAP-21050 -->
```
tivv: Please mark it as TODO

tivv (approved):
LGTM with minor comments

PR #315, opened 2025-08-20
Removing snapshot for CDAP 6.11.1 release

123-komal:


cdap-authorization-dataset-extn/pom.xml
```diff
@@ -22,7 +22,7 @@
   <parent>
     <groupId>io.cdap.cdap</groupId>
     <artifactId>cdap-security-extensions</artifactId>
-    <version>0.6.0-SNAPSHOT</version>
+    <version>0.6.0</version>
```
vsethi09, line 25: This should be `0.20.0`

See: https://github.com/cdapio/cdap-security-extn/pull/304/files

